PR #6, opened 2026-01-25
feat(cli): add support for snapshots

QuiiBz:
Add support for snapshots:

```
sandbox snapshot --stop <sandbox-id>
sandbox create --snapshot <snapshot-id>
sandbox snapshots list
sandbox snapshots delete <snapshot-id>
```

### DIFF
--- a/packages/sandbox/src/commands/snapshot.ts
+++ b/packages/sandbox/src/commands/snapshot.ts
@@ -67,12 +67,6 @@ export const snapshot = cmd.command({
 
     const spinner = silent ? undefined : ora("Creating snapshot...").start();
     const snapshot = await sandbox.snapshot();
-    spinner?.stop();
-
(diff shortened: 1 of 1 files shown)


packages/sandbox/src/commands/snapshot.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const spinner = silent ? undefined : ora("Creating snapshot...").start();
+    const snapshot = await sandbox.snapshot();
+    spinner?.stop();
+
+    if (!silent) {
+      process.stderr.write("✅ Snapshot ");
+      process.stdout.write(chalk.cyan(snapshot.snapshotId));
+      process.stderr.write(" created.\n");
+    }
```
Schniz: nit: this can probably be `spinner?.succeed(...)`

PR #16, opened 2026-01-28
Update sandbox list command to include pending sandboxes by default

Timer:
Always show pending Sandboxes, and give a hint for an empty list.


JohnPhamous (approved):

QuiiBz (changes requested):
Can you add a changeset?

PR #20, opened 2026-01-28
feat(network policy): add network policy support

valerian-roche:
This PR adds CLI support for network policies in sandboxes:
```
 % ./bin/sandbox.mjs create --network-policy no-access
✅ Sandbox sbx_32LpqfJr3Y4YJiegTdGWt10ghg81 created.
   │ team: valerian-roches-projects
   ╰ project: vercel-sandbox-default-project
   
  % sandbox ssh sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
▲ /vercel/sandbox/ curl google.com
curl: (6) Could not resolve host: google.com

 % ./bin/sandbox.mjs config network-policy --network-policy internet-access
 ✅ Network policy updated for sandbox sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
   ╰ mode: internet-access
   
 % sandbox ssh sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
 ▲ /vercel/sandbox/ curl google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">

 % ./bin/sandbox.mjs config network-policy sbx_32LpqfJr3Y4YJiegTdGWt10ghg81 --network-policy restricted --allowed-domain "*.google.com" --allowed-domain "google.com"
✅ Network policy updated for sandbox sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
   ╰ mode: restricted

 % sandbox ssh sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
▲ /vercel/sandbox/ curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">

# Should fail, missing hive support in spark
▲ /vercel/sandbox/ curl https://amazon.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">

# Allow google CIDRs only
 % ./bin/sandbox.mjs config network-policy sbx_32LpqfJr3Y4YJiegTdGWt10ghg81 --network-policy restricted --allowed-cidr  "142.0.0.0/8" --allowed-cidr "172.0.0
.0/8"
✅ Network policy updated for sandbox sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
   ╰ mode: restricted
   
 % sandbox ssh sbx_32LpqfJr3Y4YJiegTdGWt10ghg81
▲ /vercel/sandbox/ curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">

▲ /vercel/sandbox/ curl -v https://www.amazon.com
* Host www.amazon.com:443 was resolved.
* IPv4: 3.170.42.148
*   Trying 3.170.42.148:443...
* connect to 3.170.42.148 port 443 from 192.168.7.227 port 51028 failed: No route to host
```

It adds:
 - in the `create` command, adds optional network-policy flags. By default it will create a sandbox with an unrestricted "default-allow" internet policy
 - a new `config network policy` command to update a live sandbox

### DIFF
--- a/packages/sandbox/src/commands/config.ts
+++ b/packages/sandbox/src/commands/config.ts
@@ -69,7 +69,7 @@ This is a full update, fully overriding the pre-existing configuration.`,
 
       process.stderr.write(
         "✅ Network policy updated for sandbox " +
-          chalk.cyan(sandboxId) +
+          chalk.cyan(sandbox.sandboxId) +
           "\n",
       );
       process.stderr.write(


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -606,6 +619,40 @@ export class Sandbox {
...
+   * await sandbox.updateNetworkPolicy({
+   *   type: "restricted",
+   *   allowedDomains: ["*.npmjs.org", "github.com"],
+   * });
+   *
+   * @example
+   * // Deny all network access
+   * await sandbox.updateNetworkPolicy({ type: "no-access" });
+   */
+  async updateNetworkPolicy(
```
ctgowrie, line 641: Do we need to expose update network policy? I'd expect these to be immutable within the lifetime of the sandbox. Great if we support it from an infra perspective, but I'd avoid adding API surface area unless we have an explicit customer requirement.

packages/sandbox/src/commands/config.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    const spinner = ora("Updating network policy...").start();
+    try {
+      await sandbox.updateNetworkPolicy(networkPolicy);
+      spinner.stop();
+
+      process.stderr.write(
+        "✅ Network policy updated for sandbox " +
+          chalk.cyan(sandboxId) +
```
robherley:
```suggestion
          chalk.cyan(sandbox.sandboxId) +
```

because sandboxId can be a `string | Sandbox`?

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: My understanding is that it was requested by users. The main use case would be:
 - start my sandbox, download the world
 - start my agent doing yolo, I want a locked down sandbox
Users could leverage snapshots, but it's much faster to do it this way. From previous job this was one of the first request for agent sandbox to be able to allow/deny/allow (to push results)

packages/sandbox/src/commands/config.ts
valerian-roche: Updated

packages/vercel-sandbox/src/sandbox.ts
ctgowrie: Makes sense 👍

PR #33, opened 2026-02-03
feat(network firewall): align to new naming convention for network policies

valerian-roche:
In order to simplify comprehension, and towards future evolution, this commit updates the network policy naming convention:
 - basic modes `allow-all` (default) and `deny-all` are made simpler, as top-level definitions
 - `internet-access` is not `allow-all`, making it more explicit that it might allow private network in the future
 - `no-access` is now `deny-all`, as a direct mirror of the renaming of `internet-access`. It also makes it more obvious that it blocks everything, including DNS and ICMP
 - `restricted` is now simply a structure with restrictions. It keeps the functional meaning of "denied by default"

### DIFF
--- a/packages/vercel-sandbox/src/api-client/api-client.ts
+++ b/packages/vercel-sandbox/src/api-client/api-client.ts
@@ -576,7 +576,7 @@ export class APIClient extends BaseClient {
         signal: params.signal,
       }),
     );
-    return fromAPINetworkPolicy(response.json.networkPolicy);
+    return fromAPINetworkPolicy(response.json.sandbox.networkPolicy!);
   }
 
   async extendTimeout(params: {
--- a/packages/vercel-sandbox/src/api-client/validators.ts
+++ b/packages/vercel-sandbox/src/api-client/validators.ts
@@ -2,6 +2,19 @@ import { z } from "zod";
 
 export type SandboxMetaData = z.infer<typeof Sandbox>;
 
+export const NetworkPolicyValidator = z.union([
+  z.object({ mode: z.literal("allow-all") }).passthrough(),
+  z.object({ mode: z.literal("deny-all") }).passthrough(),
+  z
+    .object({
+      mode: z.literal("custom"),
+      allowedDomains: z.array(z.string()).optional(),
+      allowedCIDRs: z.array(z.string()).optional(),
+      deniedCIDRs: z.array(z.string()).optional(),
+    })
+    .passthrough(),
+]);
+
 export const Sandbox = z.object({
   id: z.string(),
   memory: z.number(),
@@ -28,6 +41,7 @@ export const Sandbox = z.object({
   cwd: z.string(),
   updatedAt: z.number(),
   interactivePort: z.number().optional(),
+  networkPolicy: NetworkPolicyValidator.optional(),
 });
 
 export type SandboxRouteData = z.infer<typeof SandboxRoute>;
@@ -141,22 +155,8 @@ export const ExtendTimeoutResponse = z.object({
   sandbox: Sandbox,
 });
 
-export const NetworkPolicyValidator = z.union([
-  z.object({ mode: z.literal("allow-all") }).passthrough(),
-  z.object({ mode: z.literal("deny-all") }).passthrough(),
-  z
-    .object({
-      mode: z.literal("custom"),
-      allowedDomains: z.array(z.string()).optional(),
-      allowedCIDRs: z.array(z.string()).optional(),
-      deniedCIDRs: z.array(z.string()).optional(),
-    })
-    .passthrough(),
-]);
-
 export const UpdateNetworkPolicyResponse = z.object({
   sandbox: Sandbox,
-  networkPolicy: NetworkPolicyValidator,
 });
 
 export const CreateSnapshotResponse = z.object({


packages/sandbox/src/util/network-policy.ts
```diff
@@ -1,7 +1,7 @@
 import chalk from "chalk";
 import type { NetworkPolicy } from "@vercel/sandbox";
 
-type NetworkPolicyMode = "internet-access" | "no-access" | "restricted";
+type NetworkPolicyMode = "allow-all" | "deny-all" | "custom";
```
robherley: To avoid a breaking change, should we keep the legacy options?

```suggestion
type DeprecatedNetworkPolicyMode = "internet-access" | "no-access" | "restricted";
type NetworkPolicyMode = "allow-all" | "deny-all" | "custom" | DeprecatedNetworkPolicyMode;
```

packages/vercel-sandbox/src/utils/network-policy.ts
```diff
@@ -58,12 +58,12 @@ export function toAPINetworkPolicy(
   }
 
   const { type, ...rest } = policy;
```
robherley:
```suggestion
  const { mode, ...rest } = policy;
```

packages/vercel-sandbox/src/utils/network-policy.ts
```diff
@@ -58,12 +58,12 @@ export function toAPINetworkPolicy(
...
-  switch (policy.type) {
-    case "internet-access":
+  switch (policy.mode) {
```
robherley:
```suggestion
  switch (mode) {
```

packages/vercel-sandbox/src/utils/network-policy.ts
valerian-roche: Updated

packages/vercel-sandbox/src/utils/network-policy.ts
valerian-roche: No longer applicable

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -630,27 +629,25 @@ export class Sandbox {
...
-      signal: opts?.signal,
-    });
+  ): Promise<NetworkPolicy> {
+    return this.client
+      .updateNetworkPolicy({
+        sandboxId: this.sandbox.id,
+        networkPolicy: networkPolicy,
+        signal: opts?.signal,
+      })
+      .then((response) => response.json.networkPolicy);
```
QuiiBz: We should also store `networkPolicy` on the `SandboxMetadata` object so we can access it (e.g. similar to `timeout`), and update it within `updateNetworkPolicy()` (e.g. similar to `extendTimeout` https://github.com/vercel/sandbox/blob/5913293e31841bc790ef86859f71fa3f8a1f0b91/packages/vercel-sandbox/src/sandbox.ts#L683)

QuiiBz (approved):

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: I need Rob's PR to be done first. Can either wait for it or do it in follow-up

packages/sandbox/docs/index.md
```diff
@@ -52,20 +52,20 @@ OPTIONS:
   --timeout <num UNIT>              - The maximum duration a sandbox can run for. Example: 5m, 1h [default: 5 minutes]
   --publish-port <PORT>, -p=<PORT>  - Publish sandbox port(s) to DOMAIN.vercel.run
   --snapshot, -s <snapshot_id>      - Start the sandbox from a snapshot ID [optional]
-  --network-policy <MODE>           - Network policy mode: "internet-access", "no-access", or "restricted"
+  --network-policy <MODE>           - Network policy mode: "allow-all", "deny-all", or "custom"
```
robherley: I think these docs need to be updated, we need to drop `custom` since it's inferred

packages/sandbox/src/commands/config.ts
```diff
@@ -64,16 +76,18 @@ This is a full update, fully overriding the pre-existing configuration.`,
...
+      const response = await sandbox.updateNetworkPolicy(networkPolicy);
       spinner.stop();
 
       process.stderr.write(
         "✅ Network policy updated for sandbox " +
           chalk.cyan(sandbox.sandboxId) +
           "\n",
       );
+      const mode =
+        typeof response === "string" ? response : "restricted";
```
robherley, line 88: I don't think this should be `restricted`

.changeset/itchy-items-dig.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/sandbox": minor
+"sandbox": minor
+---
+
+Use new model for network-polices
```
robherley:
```suggestion
Use new model for network-policies
```

packages/vercel-sandbox/src/sandbox.ts
robherley: I can do it in a follow up

packages/sandbox/src/commands/config.ts
```diff
@@ -3,33 +3,45 @@ import { Sandbox } from "@vercel/sandbox";
...
+import { buildNetworkPolicy, resolveMode } from "../util/network-policy";
 import ora from "ora";
 import chalk from "chalk";
 
 const networkPolicyCommand = cmd.command({
   name: "network-policy",
   description: `Update the network policy of a sandbox
-  
-This is a full update, fully overriding the pre-existing configuration.`,
+  This is a full update, fully overriding the pre-existing configuration.`,
```
QuiiBz: Missing a `.` for the first sentence, and the second one is maybe too verbose?
```suggestion
  description: `Update the network policy of a sandbox.
This will override the previous configuration.`,
```

packages/vercel-sandbox/src/sandbox.ts
robherley: Actually this might be tricky. 

@valerian-roche we're returning the full sandbox in the response (which includes the policy) instead of the policy in the top level object. We might merge the API PR first and update this.

valerian-roche: Go ahead merging API so I can adapt

packages/sandbox/docs/index.md
valerian-roche: Agh missed that we duplicate this. Seems very brittle

packages/sandbox/src/commands/config.ts
valerian-roche: Imo it's better than custom. I asked Claude for a new name and it came back with `restricted` also

.changeset/itchy-items-dig.md
valerian-roche: Updated

packages/sandbox/src/commands/config.ts
valerian-roche: Updated

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -576,7 +576,7 @@ export class APIClient extends BaseClient {
         signal: params.signal,
       }),
     );
-    return fromAPINetworkPolicy(response.json.networkPolicy);
+    return fromAPINetworkPolicy(response.json.sandbox.networkPolicy!);
```
robherley, line 579: It doesn't need to happen in this PR, but similar to @QuiiBz 's comment, we should return the full sandbox here and replace the object metadata.

I think it's fine for now, since we're not exposing the network policy on the sandbox (yet).

valerian-roche: Agh this got lost in a refactor. I actually don't like the setup on extend timeout as it makes the function synchronous

PR #34, opened 2026-02-05
Use @vercel/oidc utilities to reduce auth duplication

gr2m:
## Summary

This PR refactors the authentication logic to use utilities from `@vercel/oidc`, reducing code duplication and improving maintainability.

## Changes

- Use `@vercel/oidc` optional parameters to reduce duplication in auth flow
- Remove duplicate auth logic by leveraging shared utilities
- Update dependency to local path for testing purposes
- Update tests to reflect simplified auth logic

## Related

Depends on vercel/vercel#14864 which adds the utilities to `@vercel/oidc` package.

## Test plan

- [x] Existing auth tests pass
- [x] Manual testing of auth flows
- [x] Verify behavior matches previous implementation

### DIFF
--- a/.changeset/vercel-oidc-update.md
+++ b/.changeset/vercel-oidc-update.md
@@ -1,0 +1,6 @@
+---
+"sandbox": patch
+"@vercel/sandbox": patch
+---
+
+Update to use `@vercel/oidc@3.2.0` utilities, removing duplicate auth logic and the local `JwtExpiry` class


packages/sandbox/src/args/auth.ts
```diff
@@ -38,79 +37,46 @@ export const token = cmd.option({
...
-      if (auth) {
-        const refreshed = await refreshToken(auth);
-        if (typeof refreshed === "object") {
-          auth = refreshed;
-        } else if (
-          refreshed === "missing refresh token" ||
-          refreshed === "invalid refresh token"
+      // Try to get CLI token, which handles auth.json reading and refresh
+      try {
+        return await getVercelCliToken();
```
gr2m: Do we need an access token? Or could we also get an OIDC token for Sandbox?

dvoytenko: Handling the access token is good idea. But then it should actually be called "access token". And we should also allow it be specified via an env var, e.g. `VERCEL_TOKEN`. But IMHO it shouldn't be API'd to CLI specifically.

packages/sandbox/package.json
```diff
@@ -46,7 +46,7 @@
     "@types/debug": "^4.1.12",
     "@types/ms": "^2.1.0",
     "@types/node": "^22.15.12",
-    "@vercel/oidc": "^3.1.0",
+    "@vercel/oidc": "file:../../../vercel/packages/oidc",
```
dvoytenko: ?

gr2m: I'm working on it locally, this is a local install path for local testing. Once the `@vercel/oidc` PR ships I'll update it.

packages/vercel-sandbox/package.json
```diff
@@ -30,7 +30,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@vercel/oidc": "^3.1.0",
+    "@vercel/oidc": "3.2.0",
```
gr2m, line 33: let me know if you want a range, but based on past experience I'd recommend to pin it

packages/sandbox/src/args/auth.ts
```diff
@@ -27,6 +26,10 @@ export const token = cmd.option({
 
       if (process.env.VERCEL_OIDC_TOKEN) {
         try {
+          // Note: getVercelOidcToken() can optionally accept { project, team } parameters
+          // to explicitly specify the scope. `project` must be a project ID (prj_*),
+          // `team` accepts both IDs (team_*) and slugs. If not provided, it will
+          // read from .vercel/project.json automatically.
```
gr2m:
```suggestion
```

packages/sandbox/src/util/infer-scope.ts
```diff
@@ -5,10 +5,17 @@ import * as Auth from "@vercel/sandbox/dist/auth/index.js";
...
 export type InferredScope = {
+  /** Project ID (prj_*) or slug */
   projectId: string;
+  /** Team/owner ID (team_*) or slug */
   ownerId: string;
+  /** Human-readable project slug for display */
   projectSlug?: string;
+  /** Human-readable team slug for display */
   ownerSlug?: string;
 };
```
gr2m, line 20: let me clean this up

packages/vercel-sandbox/src/utils/get-credentials.ts
```diff
@@ -61,7 +61,14 @@ async function getVercelToken(opts: {
   projectId?: string;
 }): Promise<Credentials> {
   try {
-    return getCredentialsFromOIDCToken(await getVercelOidcToken());
+    // Pass team and project to getVercelOidcToken to enable token refresh
+    // without needing to read from .vercel/project.json. `project` must be a
+    // project ID (prj_*), `team` accepts both IDs (team_*) and slugs.
```
gr2m:
```suggestion
```

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -59,23 +60,56 @@ export class APIClient extends BaseClient {
     });
 
     this.teamId = params.teamId;
-    this.tokenExpiry = JwtExpiry.fromToken(params.token);
+    this.isJwtToken = params.token.split(".").length === 3;
```
dvoytenko: Somewhat of a weak check. An access token can very validly use JWT format, but not be an OIDC token. Ideally we'd try to parse as our OIDC and check that it actually has the confirming claims and if not, we set `this.isJwtToken` to `false`.

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -59,23 +60,57 @@ export class APIClient extends BaseClient {
...
+        // Verify this is actually a Vercel OIDC token by checking for owner_id claim
+        if (payload.owner_id) {
+          this.isJwtToken = true;
+          this.projectId = payload.project_id;
+          this.teamId = payload.owner_id;
+        }
+      } catch {
+        // Not a valid OIDC token, keep isJwtToken as false
+      }
+    }
```
Schniz: nit: let's extract this to `decodeUnverifiedToken(token): Claims | null`

Schniz: seems like this function will be used in `ensureValidToken` too <picture data-single-emoji=":relieved-telegram:" title=":relieved-telegram:">:relieved-telegram:</picture>

Schniz (approved):

Schniz (approved):

PR #36, opened 2026-02-06
feat(sdk,cli): support custom/infinite snapshots expiration

QuiiBz:
Add support for custom or infinite snapshots expiration, instead of the current 7 days default. If unspecified, we'll still fallback to the default expiration

Similar to the `timeout` field in `Sandbox.create()`, we accept a new `expiration` field in `sandbox.snapshot()` that represents the desired snapshot expiration in milliseconds. If set to 0, the snapshot will never expire

### DIFF
--- a/packages/sandbox/src/commands/snapshot.ts
+++ b/packages/sandbox/src/commands/snapshot.ts
@@ -21,7 +21,7 @@ export const args = {
   expiration: cmd.option({
     long: "expiration",
     type: cmd.optional(Duration),
-    description: "The expiration time of the snapshot. Use 0 for no expiration at all.",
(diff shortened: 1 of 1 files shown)


packages/sandbox/src/commands/snapshot.ts
```diff
@@ -16,6 +18,11 @@ export const args = {
     long: "silent",
     description: "Don't write snapshot ID to stdout",
   }),
+  expiration: cmd.option({
+    long: "expiration",
+    type: cmd.optional(Duration),
+    description: "The expiration time of the snapshot. Use 0 for no expiration at all.",
```
AndyW22:
```suggestion
    description: "The expiration time of the snapshot. Use 0 for no expiration.",
```

PR #41, opened 2026-02-09
feat(network firewall): provide policy getter on Sandbox instance

valerian-roche:
# Summary
The new `network policy` attribute is currently not accessible by the user on the Sandbox object.
This change adds an accessor for the property that is otherwise already returned by the `updateNetworkPolicy` method

### DIFF
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -15,7 +15,7 @@ import { consumeReadable } from "./utils/consume-readable";
 import {
   type NetworkPolicy,
 } from "./network-policy";
-import { fromAPINetworkPolicy } from "./utils/network-policy";
+import { convertSandbox, type ConvertedSandbox } from "./utils/convert-sandbox";
 
 export type { NetworkPolicy };
 
@@ -203,9 +203,7 @@ export class Sandbox {
   /**
    * Internal metadata about this sandbox.
    */
-  private sandbox: Omit<SandboxMetaData, "networkPolicy"> & {
-    networkPolicy?: NetworkPolicy;
-  };
+  private sandbox: ConvertedSandbox;
 
   /**
    * Allow to get a list of sandboxes for a team narrowed to the given params.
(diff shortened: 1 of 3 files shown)


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -682,7 +704,7 @@ export class Sandbox {
     });
 
     // Update the internal sandbox metadata with the new timeout value
-    this.sandbox = response.json.sandbox;
+    this.sandbox.timeout = response.json.sandbox.timeout;
```
QuiiBz: Not sure about this change, it wouldn't pick up changes to e.g. `updatedAt`

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -302,6 +312,16 @@ export class Sandbox {
    * @param routes - Port-to-subdomain mappings for exposed ports
    * @param sandboxId - Unique identifier for the sandbox
    */
+  private static convertSandbox(sandbox: SandboxMetaData): Sandbox["sandbox"] {
```
QuiiBz: This will break the JSDoc of the constructor, can we move that function somewhere else?

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: Okay then I need to do the reverse in updateNetworkPolicy. Will change

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: I can move the function, but for my information, why is a private function impacting JSDoc?

valerian-roche: Updated

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: Updated

packages/vercel-sandbox/src/sandbox.ts
QuiiBz: It's not specifically due to being a private function, just that the JSDoc applies to the function just under the comment

PR #52, opened 2026-02-19
feat: support privateParams in all the endpoints

marc-vercel:
Extend the use of `privateParams` to all the other API endpoints. The `privateParams` was already supported during the `createSandbox`, what we are doing here is just:
- Storing the private parameters in a property.
- Use the same private parameters in any API call from the same `sandbox` instance.

### DIFF
--- a/packages/vercel-sandbox/src/command.ts
+++ b/packages/vercel-sandbox/src/command.ts
@@ -32,7 +32,7 @@ export class Command {
   private cmd: CommandData;
 
   /**
-   * Private parameters (e.g., dev hive params) to pass to API calls.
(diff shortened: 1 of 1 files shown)


packages/vercel-sandbox/src/command.ts
```diff
@@ -31,6 +31,11 @@ export class Command {
    */
   private cmd: CommandData;
 
+  /**
+   * Private parameters (e.g., dev hive params) to pass to API calls.
```
QuiiBz:
```suggestion
   * Private parameters to pass to API calls.
```

PR #54, opened 2026-02-20
feat(resources): add stopped sandboxes used resources to SDK and add blocking stop method

valerian-roche:
Stopped resources are returned with their billable resource usage. This commit adds SDK support for them:
 - return CPU and network usage for stopped sandboxes
 - add blocking option in the stop method It also adds resource usage in `sandbox list` command if requested to show stopped sandboxes

```typescript
const sandbox = await Sandbox.create();
const result = await sandbox.runCommand("curl", [
  "-s",
  "https://httpbin.org/get",
]);
console.log(await result.stdout());

const stopped = await sandbox.stop({ blocking: true });
console.log(`Sandbox ${stopped.id} [${stopped.status}]`);
console.log(`  CPU: ${stopped.activeCpuDurationMs ?? 0}ms`);
console.log(
  `  Network: ${stopped.networkTransfer?.egress ?? 0}B out / ${stopped.networkTransfer?.ingress ?? 0}B in`,
);

```

<details>
<summary>Result</summary>

```
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.org",
    "User-Agent": "curl/8.15.0",
    "X-Amzn-Trace-Id": "Root=1-6998f305-2470ff825861636c13629dc5"
  },
  "origin": "18.207.212.203",
  "url": "https://httpbin.org/get"
}

Sandbox sbx_DrOByIoQ567npaUS3Gl3mXAZmv2z [stopped]
  CPU: 1749ms
  Network: 7727B out / 11111B in
```

</details>

### DIFF
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -204,6 +204,20 @@ export class Sandbox {
   }
 
   /**
+   * The amount of CPU used by the sandbox. Only reported once the VM is stopped.
+   */
+  public get activeCpuUsageMs(): number | undefined {
+    return this.sandbox.activeCpuDurationMs;
+  }
+
+  /**
+   * The amount of network data used by the sandbox. Only reported once the VM is stopped.
+   */
+  public get networkTransfer(): {ingress: number, egress: number} | undefined {
+    return this.sandbox.networkTransfer;
+  }
+
+  /**
    * Internal metadata about this sandbox.
    */
   private sandbox: ConvertedSandbox;


packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -609,18 +609,37 @@ export class APIClient extends BaseClient {
...
       SandboxResponse,
       await this.request(url, {
         method: "POST",
         ...(hasPrivateParams && { body: JSON.stringify(privateParams) }),
         signal: params.signal,
       }),
     );
+
+    if (params.blocking) {
+      const { setTimeout } = await import("node:timers/promises");
```
QuiiBz: Can we move this import to the top-level? No need to lazy-load

valerian-roche: Updated

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -203,6 +203,20 @@ export class Sandbox {
     return this.sandbox.sourceSnapshotId;
   }
 
+  /**
```
valerian-roche, line 206: Also added those. I don't understand if they're needed or not

QuiiBz: 👍 it's needed because `this.sandbox` is private

PR #108, opened 2026-03-20
feat(vercel-sandbox): support onResume property for Get

marc-vercel:
Add a new optional property, `onResume`, to the `Sandbox.get()`. When set:

- Automatically sets `resume` to true.
- Automatically runs the provided command in `onResume` in **detached mode**. This means we do not wait until its completion.


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -421,21 +426,32 @@ export class Sandbox {
     });
 
     const privateParams = getPrivateParams(params);
+    const resume = !!params.onResume ? true : params.resume;
```
LukeSheard:
```suggestion
    const resume = Boolean(params.resume || params.onResume);
```

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -121,6 +121,11 @@ interface GetSandboxParams {
    * Whether to resume an existing session. Defaults to true.
    */
   resume?: boolean;
+  /**
+   * A command to execute after the sandbox is resumed.
+   * When set, `resume` is automatically set to `true`.
+   */
+  onResume?: { command: string; args?: string[] };
```
LukeSheard:
```suggestion
  onResume?: RunCommandParams;
```

Probably want to expose the whole command option instead of a subset?

PR #112, opened 2026-03-25
Add `node:fs` implementation on `Sandbox`

LukeSheard:
This PR introduces a node:fs-compatible API for the Vercel Sandbox environment.

The goal is to provide a familiar, minimal subset of Node’s filesystem interface that works within the constraints of the Sandbox runtime, enabling easier portability of existing Node.js code and libraries.

In particular provide something we can pass directly around with `just-bash` as an `fs` implementation without needing their `Sandbox` primitive wrapper.

### DIFF
--- a/packages/vercel-sandbox/src/filesystem.ts
+++ b/packages/vercel-sandbox/src/filesystem.ts
@@ -1,6 +1,19 @@
-import * as fs from "fs";
+import type { Stats, Dirent } from "fs";
+import * as constants from "node:constants";
 
-// UV_DIRENT_* constants exist at runtime but aren't in @types/node
+const {
+  S_IFMT,
+  S_IFREG,
+  S_IFDIR,
+  S_IFLNK,
+  S_IFBLK,
+  S_IFCHR,
+  S_IFIFO,
+  S_IFSOCK,
+} = constants;
+
+// https://github.com/nodejs/node/blob/main/deps/uv/include/uv.h#L1267-L1276
+// exposed through node:constants but not in @types/node
 const UV_DIRENT_FILE = 1;
 const UV_DIRENT_DIR = 2;
 const UV_DIRENT_LINK = 3;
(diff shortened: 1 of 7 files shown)


packages/vercel-sandbox/src/filesystem.ts
```diff
@@ -0,0 +1,712 @@
...
+ * Access via {@link Sandbox.fs}.
+ *
+ * @hideconstructor
+ */
+
+import { Stats, Dirent } from "fs";
+
+// The Stats constructor and Dirent constructor exist at runtime but are marked
+// private in @types/node. We cast to the actual runtime signatures.
+const StatsConstructor = Stats as unknown as new (
```
QuiiBz: Ouch, there must a reason why it's marked private no?

packages/vercel-sandbox/src/filesystem.ts
```diff
@@ -0,0 +1,712 @@
...
+    syscall: string;
+    path: string;
+  };
+  err.code = code;
+  err.syscall = syscall;
+  err.path = path;
+  return err;
+}
+
+interface SandboxHandle {
```
QuiiBz, line 62: Can we do a `Pick` to avoid having to update both definitions if we update these methods?

```suggestion
type SandboxHandleTest = Pick<SandboxHandle, "readFileToBuffer" | "writeFiles" | "mkDir" | "runCommand">;
```

LukeSheard:
* **#112** <a href="https://app.graphite.com/github/pr/vercel/sandbox/112?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/sandbox/112?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/vercel-sandbox/src/filesystem.ts
```diff
@@ -0,0 +1,712 @@
...
+) => Stats;
+
+const DirentConstructor = Dirent as unknown as new (
+  name: string,
+  type: number,
+  path: string,
+) => Dirent;
+
+// UV_DIRENT_* constants exist at runtime but aren't in @types/node
+const UV_DIRENT_FILE = 1;
```
QuiiBz, line 17: Can we add a comment to link to the Node.js source every time we're extracing logic like this?

QuiiBz (changes requested):
We also need a changeset

packages/vercel-sandbox/src/filesystem.ts
LukeSheard: I've updated this to just pass through the entire thing now.

packages/vercel-sandbox/src/filesystem.ts
LukeSheard: Done

packages/vercel-sandbox/src/filesystem.ts
LukeSheard: Yeah, I've created my own version for the binding that satisfies the interface.

QuiiBz (approved):

PR #120, opened 2026-03-28
fix: smarter fallback team selection for scope inference

gr2m:
## Summary

- Fetches both `/v2/user` and `/v2/teams` in parallel to build an ordered list of candidate teams for scope inference
- Tries `user.defaultTeamId` first, then falls back to the best hobby-plan OWNER team (personal team matching username, or most recently updated)
- Filters fallback candidates by `billing.plan === 'hobby'` and `membership.role === 'OWNER'` to avoid selecting pro/enterprise teams
- On 403, skips to the next candidate instead of failing immediately
- Only throws after all candidates are exhausted, with a helpful error message including the authenticated username

### DIFF
--- a/.changeset/slick-lands-repeat.md
+++ b/.changeset/slick-lands-repeat.md
@@ -2,4 +2,4 @@
 "@vercel/sandbox": patch
 ---
 
-Smarter fallback team selection for scope inference: tries `defaultTeamId` first, then the best OWNER team (personal team or most recently updated). Skips teams that return 403 and shows a helpful error when no team allows sandbox creation.
+Smarter fallback team selection for scope inference: tries `defaultTeamId` first, then the best hobby-plan OWNER team (personal team or most recently updated). Filters fallback candidates by `billing.plan === 'hobby'` to avoid selecting pro/enterprise teams. Skips teams that return 403 and shows a helpful error when no team allows sandbox creation.
--- a/.changeset/slick-lands-repeat.md
+++ b/.changeset/slick-lands-repeat.md
@@ -1,5 +1,6 @@
 ---
 "@vercel/sandbox": patch
+"@vercel/vercel-sandbox": patch
 ---
 
 Smarter fallback team selection for scope inference: tries `defaultTeamId` first, then the best hobby-plan OWNER team (personal team or most recently updated). Filters fallback candidates by `billing.plan === 'hobby'` to avoid selecting pro/enterprise teams. Skips teams that return 403 and shows a helpful error when no team allows sandbox creation.
--- a/.changeset/slick-lands-repeat.md
+++ b/.changeset/slick-lands-repeat.md
@@ -2,4 +2,4 @@
 "@vercel/sandbox": patch
 ---
 
-Use the user's default team (`defaultTeamId`) for scope inference fallback instead of picking the first team from `/v2/teams`.
+Smarter fallback team selection for scope inference: tries `defaultTeamId` first, then the best OWNER team (personal team or most recently updated). Skips teams that return 403 and shows a helpful error when no team allows sandbox creation.
--- a/.changeset/slick-lands-repeat.md
+++ b/.changeset/slick-lands-repeat.md
@@ -1,0 +1,5 @@
+---
+"@vercel/sandbox": patch
+---
+
+Use the user's default team (`defaultTeamId`) for scope inference fallback instead of picking the first team from `/v2/teams`.
--- a/packages/vercel-sandbox/src/auth/project.ts
+++ b/packages/vercel-sandbox/src/auth/project.ts
@@ -32,6 +32,11 @@ const TeamsSchema = z.object({
 
 const DEFAULT_PROJECT_NAME = "vercel-sandbox-default-project";
 
+/** Status codes that mean "this team can't be used, try the next one". */
+function isSkippableTeamError(e: unknown): boolean {
+  return e instanceof NotOk && (e.response.statusCode === 402 || e.response.statusCode === 403);
+}
+
 /**
  * Resolves the team and project scope for sandbox operations.
  *
@@ -103,7 +108,7 @@ export async function inferScope(opts: {
         return result;
       }
     } catch (e) {
-      if (!(e instanceof NotOk) || e.response.statusCode !== 403) throw e;
+      if (!isSkippableTeamError(e)) throw e;
     }
   }
 
@@ -136,7 +141,7 @@ export async function inferScope(opts: {
         const result = await tryTeam(opts.token, bestHobbyTeam.id);
         return { ...result, teamSlug: bestHobbyTeam.slug };
       } catch (e) {
-        if (!(e instanceof NotOk) || e.response.statusCode !== 403) throw e;
+        if (!isSkippableTeamError(e)) throw e;
       }
     }
   } while (next !== null);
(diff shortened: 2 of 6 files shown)


packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -21,16 +18,17 @@ const DEFAULT_PROJECT_NAME = "vercel-sandbox-default-project";
...
  *
  * @param opts.token - Vercel API authentication token.
- * @param opts.teamId - Optional team slug. If omitted, the first team is selected.
+ * @param opts.teamId - Optional team slug. If omitted, the user's default team is used.
  * @param opts.cwd - Optional directory to search for `.vercel/project.json`. Defaults to `process.cwd()`.
  * @returns The resolved scope with `projectId`, `teamId`, and whether the project was `created`.
  *
  * @throws {NotOk} If the API returns an error other than 404 when checking the project.
- * @throws {ZodError} If no teams exist for the account.
+ * @throws {ZodError} If the user API response is missing a username.
```
gr2m: I don't think that's possible


```suggestion
```

packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -48,12 +59,44 @@ export async function inferScope(opts: {
...
+ */
+async function tryTeam(
+  token: string,
+  teamId: string,
+): Promise<{ projectId: string; teamId: string; created: boolean }> {
   let created = false;
   try {
     await fetchApi({
-      token: opts.token,
+      token,
```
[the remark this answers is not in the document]
gr2m: I think both should work but will double check with accounts team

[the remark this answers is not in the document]
gr2m: Yes both work, tested with 

```
sandbox create --scope gr2m
sandbox create --scope team_iCHk38bwGlsMhugWzQv5N8ea
```

packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -48,22 +63,58 @@ export async function inferScope(opts: {
...
+      if (e instanceof NotOk && e.response.statusCode === 403) {
+        continue;
+      }
+      throw e;
+    }
+  }
+
+  throw new NotOk({
+    statusCode: 403,
+    responseText: `Authenticated as "${username}" but none of the available teams allow sandbox creation. Specify a team explicitly with --scope <team-id-or-slug>.`,
```
marc-vercel, line 159: Not a blocker: it would be nice to know which teams have we tried.

packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -76,17 +127,50 @@ export async function inferScope(opts: {
...
-  const {
-    teams: [team],
-  } = await fetchApi({ token, endpoint: "/v2/teams?limit=1" }).then(
-    TeamsSchema.parse,
+export async function selectTeams(
+  token: string,
+): Promise<{ candidateTeamIds: string[]; username: string }> {
+  const [userData, teamsData] = await Promise.all([
+    fetchApi({ token, endpoint: "/v2/user" }).then(UserSchema.parse),
+    fetchApi({ token, endpoint: "/v2/teams?limit=100" }).then(
```
marc-vercel: What happens if the user has more than 100 teams? I understand we are just covering the first 100 teams.

This is unlikely, but we should iterate on that. Not a blocker.

packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -48,22 +63,58 @@ export async function inferScope(opts: {
...
   } catch (e) {
     if (!(e instanceof NotOk) || e.response.statusCode !== 404) {
       throw e;
     }
 
     await fetchApi({
-      token: opts.token,
-      endpoint: `/v11/projects?slug=${encodeURIComponent(teamId)}`,
+      token,
+      endpoint: `/v11/projects?${teamParam}`,
```
marc-vercel, line 191: Does this API endpoint ensures that the user can read sandboxes? Afaik it just checks that the user can read the project no?

For example, we require the `Resource.VercelSandbox` scope at the API level to read a sandbox.

marc-vercel (approved):
Overall LGTM, left 3 comments (two of them are optional).

I see that the tests mock the API response, but given that we are changing API requests/responses, have we manually tested all the flows? If possible I would love to have some integration tests, but because we are dealing with auth I do not think it will be easy.

packages/vercel-sandbox/src/auth/project.ts
QuiiBz: Fetching all the teams (up to 100) will be quite slow right? In which case would a user not have a `defaultTeamId`?

gr2m: `defaultTeamId` is not always set. But we can request them sequentially

Ideally the `/v2/teams` endpoint would have a access/permission based filter and sort argument, then we'd only need to get one.

packages/vercel-sandbox/src/auth/project.ts
gr2m: best we can do right now is to filter by OWNER roles

https://github.com/vercel/sandbox/blob/e3e173a9ebb91bcfd6d36c4ee4e6d585d1478391/packages/vercel-sandbox/src/auth/project.ts#L151-L153

I don't think there is a way to retrieve a list of teams filtered by a specific permission yet

marc-vercel: So, this means we can login as a team that does not have permissions. I understand that this is already happening no? If so we can iterate on that later.

packages/vercel-sandbox/src/auth/project.ts
marc-vercel: Fixed in [35e593f](https://github.com/vercel/sandbox/pull/120/commits/35e593f6eff3eec6c9b19d1ac227a7409d1f848b) - we paginate over all the teams in batches of 20.

marc-vercel: Fixed in [35e593f](https://github.com/vercel/sandbox/pull/120/commits/35e593f6eff3eec6c9b19d1ac227a7409d1f848b) - we paginate over all the teams in batches of 20.

packages/vercel-sandbox/src/auth/project.ts
```diff
@@ -42,28 +61,129 @@ export async function inferScope(opts: {
...
+        ? "/v2/teams?limit=20"
+        : `/v2/teams?limit=20&until=${next}`;
+    const page = await fetchApi({ token: opts.token, endpoint }).then(
+      TeamsSchema.parse,
+    );
+
+    const hobbyOwnerTeams = page.teams.filter(
+      (t) => t.membership.role === "OWNER" && t.billing.plan === "hobby",
+    );
+    if (hobbyOwnerTeams.length === 0) {
```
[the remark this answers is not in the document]
marc-vercel: Fixed in [090c5b0](https://github.com/vercel/sandbox/pull/120/commits/090c5b06a03f51d6810f761c46d20db0c655f203)

packages/vercel-sandbox/src/auth/project.ts
marc-vercel: Won't do. See https://vercel.slack.com/archives/C08L9SK5SM9/p1776095807427889?thread_ts=1775655595.952329&cid=C08L9SK5SM9

.changeset/slick-lands-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/sandbox": patch
```
QuiiBz, line 2: The changeset should include `sandbox` too

gr2m:
```suggestion
"@vercel/sandbox": patch
"@vercel/vercel-sandbox": patch
```

PR #141, opened 2026-04-09
feat(vercel-sandbox): support onResume parameter in Sandbox.get and Sandbox.create

marc-vercel:
Non-breaking change. Three new integration tests have been added to ensure that the communication between SDK <-> API works properly with the new field.

Support a new parameter, `onResume`, for `Sandbox.get()` and `Sandbox.create()`. This new parameter is a hook with this format:

```
onResume: (sandbox: Sandbox) => Promise<void>
```

This hook **will only** be executed when a new session is created as part of the resume. So, for example, if I run the following code the `onResume` hook is not going to be called because we are not creating a new session (it is a no-op):

```
const sandbox = await Sandbox.create();
await Sandbox.get({ name: sandbox.name, resume: true, onResume: (() => {/* Logic here */}) });
```

But in this case, it will be called:

```
const sandbox = await Sandbox.create(onResume: (() => {/* Logic here */}) });
await sandbox.stop({ blocking: true });

// The sandbox is stopped and we need to create a new session. `onResume` hook will be called.
await sandbox.runCommand({ ... });
```


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -526,13 +542,20 @@ export class Sandbox {
...
       client,
       session: response.json.session,
       sandbox: response.json.sandbox,
       routes: response.json.routes,
       projectId: credentials.projectId,
+      onResume: params.onResume,
     });
+
+    if (response.json.resumed && params.onResume) {
+      await params.onResume(sandbox);
```
QuiiBz, line 555: Trying to think whether we should add a catch handler ourselves or let that propagate to the caller of `Sandbox.get`. I think this is fine, if `onResume` fails, it makes sense that `Sandbox.get` throws by default. In any case, the user can wrap their logic into their own catch handler

marc-vercel: I was thinking on letting it propagate rather than failing silently. The user will know if the errors can be ignored or they should fail hard on resume.

QuiiBz (approved):

PR #145, opened 2026-04-13
feat(sandbox): update API URL

QuiiBz:
Update the Vercel API URL from `vercel.com/api` to `api.vercel.com`


LukeSheard (approved):

srintern (changes requested):
will only go to sfo1

PR #147, opened 2026-04-14
feat(vercel-sandbox): support HTTP2 connection pooling

marc-vercel:
The SDK usually makes sequential requests per operation, but sometimes we do multiple requests. Now even more that we have a shared agent pool.

When one connection is held open by a streaming response, the next request must open a brand new TCP+TLS connection.

With `allowH2: true`, undici negotiates HTTP/2 on the first TLS connection. HTTP/2 multiplexes multiple request/response streams over a single TCP+TLS connection. So even while the streaming connection is in-use, we can reuse the same connection — no new handshake.

### DIFF
--- a/packages/vercel-sandbox/src/api-client/base-client.ts
+++ b/packages/vercel-sandbox/src/api-client/base-client.ts
@@ -15,6 +15,7 @@ export interface RequestParams extends RequestInit {
 
 const DEFAULT_AGENT = new Agent({
   bodyTimeout: 0, // disable body timeout to allow long logs streaming
+  allowH2: true,
 });
 
 /**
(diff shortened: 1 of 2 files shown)


packages/vercel-sandbox/src/api-client/base-client.ts
```diff
@@ -15,6 +15,7 @@ export interface RequestParams extends RequestInit {
 
 const DEFAULT_AGENT = new Agent({
   bodyTimeout: 0, // disable body timeout to allow long logs streaming
+  allowH2: true,
```
QuiiBz, line 18: We also have to set `pipelining` to reuse the same connection for concurrent requests

valerian-roche: Do they use `pipelining` to enable multiplexing? Those have very different behaviors depending on whether h1 or h2 ends up being used

PR #160, opened 2026-04-22
feat(vercel-sandbox): support Sandbox.getOrCreate method

marc-vercel:
For the beta package, introduce a new static method:

```
Sandbox.getOrCreate()
```

What it does:

1. If the sandbox has no `name`, just create it and execute the `onCreate` hook if it exists. Otherwise:
2. If the sandbox exists, just do a `Sandbox.get()` and return it.
3. If the sandbox returns `snapshot_not_found`, it means the sandbox cannot be resumed because the snapshot was deleted/expired. Delete the sandbox and create a new one with the **same name**. Execute the `onCreate` hook if it exists.
4. If the sandbox return `not_found`, create a new sandbox. Execute the `onCreate` hook if it exists.

Important: this is done at the SDK level. We might iterate in the future and implement it at the API level, which means we will need only one requests instead of two.


AndyW22 (approved):

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -571,6 +603,103 @@ export class Sandbox {
     return sandbox;
   }
 
+  /**
+   * Retrieve an existing named sandbox, or create a new one if none exists.
+   *
+   * If `name` is omitted, this always creates a new sandbox and fires
+   * `onCreate`. If `name` is provided, it first tries {@link Sandbox.get};
```
QuiiBz, line 610: Given the name of the method, `name` should be required: doesn't quite make sense to "get" if you don't provide a name

QuiiBz (changes requested):

packages/vercel-sandbox/src/sandbox.ts
marc-vercel: Great call, fixed in [494ca1d](https://github.com/vercel/sandbox/pull/160/commits/494ca1da445fc9e7d90d793aa8da01bfb57fb2d4)!

QuiiBz (approved):

PR #163, opened 2026-04-23
feat(vercel-sandbox,sandbox): support pagination when listing sandboxes, sessions, snapshots

marc-vercel:
Introduce pagination support for the CLI/SDK, so that customers and agents can iterate over existing sandboxes, snapshots and sessions.

CLI
---

The CLI now supports two new parameters to control the pagination:

- `--limit`
- `--cursor`

Also, in the list commands that support pagination, we are also returning the `More results: ...` at the end, so that humans and agents know which commands they should run next if they want to iterate over the results.

```
> sandboxes list --limit 2
...
More results: sandbox list --limit 2 --cursor eyJvZmZzZXQiOjJ9

> sandbox sessions list my-sandbox --limit 2
...
More results: sandbox sessions list my-sandbox --limit 2 --cursor eyJvZmZzZXQiOjJ9

> sandbox snapshots list --name my-sandbox --limit 2
...
More results: sandbox snapshots list --name my-sandbox --limit 2 --cursor eyJvZmZzZXQiOjJ9
```

SDK
---

SDK introduces a small breaking change for the beta customers. When executing any of these methods:

- `await Sandbox.list()`
- `await sandbox.listSessions()`
- `await sandbox.listSnapshots()`
- `await Snapshot.list()`

We are going to return an Iterable. To match the previous behavior, you will need to convert these into an array. For example:

```
const sandboxes = await (await Sandbox.list()).toArray();
```

If you want to iterate through the results and automatically let the SDK query the pages when required:

```
const result = await Sandbox.list({ namePrefix: "ci-", limit: 20 });

# Iterate item by item.
for await (const sandbox of result) {
  console.log(sandbox.name, sandbox.status);
}

# Iterate page by page.
for await (const page of result.pages()) {
  console.log(`Got ${page.sandboxes.length} sandboxes, next=${page.pagination.next}`);
}
```

To support pagination, these previous 4 methods will support two new optional parameters:

- `limit`: page size. Maximum is 100.
- `cursor`: starting point. The first page is fetched with this cursor; from there the paginator follows pagination.next automatically.


packages/sandbox/src/util/output.ts
```diff
@@ -77,4 +77,40 @@ export function formatRunDuration(d: number): string {
     return `${d}ms`;
   }
   return `${d/1000}s`
+}
+
+type FlagValue = string | number | boolean | string[] | undefined;
+
+export function formatNextCursorHint(
```
QuiiBz: Do we actually need this complex `formatNextCursorHint` logic, instead of doing something like
```ts
`More entries: ${process.argv} --cursor ${pagination.next}`
```
? Might need some extra logic to avoid injecting `--cursor` twice if already passed in, but would avoid having to manually rebuild the command and executing many regexesz

marc-vercel: Great call 💯! This simplifies a lot the code. Fixed in https://github.com/vercel/sandbox/pull/163/changes/40c444f1014e8b6a278fb8d53fa0842fc3ec8e05) (as you said I still had to add some logic, but it is minimal compared to what we had before).

QuiiBz (approved):

PR #164, opened 2026-04-23
feat(vercel-sandbox,sandbox): introduce keepLast parameters for SDK/CLI (under feature flag)

marc-vercel:
Introduce `keepLast` snapshot retention and `forceDelete` for the SDK and CLI. These changes are gated behind a feature flag until we agree that this is the correct contract.

CLI
---

`sandbox create` and the new `sandbox config keep-last` subcommand configure a retention policy that keeps only the N most recent snapshots. By default, evicted snapshots are deleted immediately; pass `--soft-evict` to let them fall back to the sandbox's default expiration instead.

```
> sandbox create --keep-last 3
> sandbox create --keep-last 3 --keep-last-for 7d
> sandbox create --keep-last 5 --keep-last-for 30d --soft-evict

> sandbox config keep-last my-sandbox 3
> sandbox config keep-last my-sandbox 3 --for 7d --soft-evict
> sandbox config keep-last my-sandbox --clear
> sandbox config list my-sandbox     # includes a new "Keep last" row
```

`sandbox snapshots delete` supports `--force` / `-f` to delete a snapshot that is currently the sandbox's ` currentSnapshotId` (server rejects the delete without it when the flag is on):

```
> sandbox snapshots delete snap_abc            # fails if the snapshot is in use
> sandbox snapshots delete snap_abc --force    # succeeds
```

SDK
---

`Sandbox.create` and `Sandbox.update` accept a new optional `snapshotKeepLast` field.

```
  snapshotKeepLast: {
    count: 3,
    expiration: 7 * 24 * 60 * 60 * 1000,   // ms, 0 = never. Falls back to snapshotExpiration when omitted.
    deleteEvicted: true,                    // default true
  },
});
```

On `Sandbox.update`, pass `null` to remove the policy:

```
await sandbox.update({ snapshotKeepLast: null });
```

`snapshot.delete` accepts `forceDelete` to bypass the server's in-use check where we fail if the latest snapshot is being used by a sandbox (and is the last snapshot):

```
await snapshot.delete({ forceDelete: true });
```


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -394,6 +414,16 @@ export class Sandbox {
     return this.sandbox.snapshotExpiration;
   }
 
+  /**
+   * The snapshot retention policy (`keep-last-snapshots`) currently configured
+   * on this sandbox, if any.
+   */
+  public get keepLastSnapshots():
+    | { count: number; expiration?: number; deleteEvicted: boolean }
```
QuiiBz, line 422: We could use `SandboxMetaData['keepLastSnapshots']` here and everywhere else where we duplicate the types definition, but I think we should instead extract into its own type (basically everything line 119-134) so we keep the TSDoc everywhere

QuiiBz (approved):

PR #171, opened 2026-04-30
feat(sdk): add L7 requests matcher for network policies

QuiiBz:
This PR adds support for L7 requests matcher in network policies: for each `NetworkPolicyRule`, you can apply a matcher on:
- the request method 
- the request path
- the request query strings
- the request headers

Each matcher can target to either exactly match, start with, or regex match, a provided value. For example, to inject an authorization header on requests to Vercel's AI gateway but only on specific requests:

```
allow: {
  "ai-gateway.vercel.sh": [
    {
      match: {
        method: ["POST"],
        path: { startsWith: "/v1/" },
        headers: [
          { key: { exact: "x-api-key" }, value: { exact: "placeholder" } }
        ]
      },
      transform: [{
        headers: { authorization: "Bearer ..." }
      }]
    }
  ],
  "*": []
}
```

### DIFF
--- a/packages/vercel-sandbox/src/network-policy.ts
+++ b/packages/vercel-sandbox/src/network-policy.ts
@@ -12,13 +12,15 @@ export type NetworkTransformer = {
 };
 
 /**
- * Defines how a request value is matched. Exactly one field should be set.
+ * Defines how a request value is matched.
  */
 export type NetworkPolicyMatcher = {
   /** Match the value exactly. */
   exact?: string;
+} | {
   /** Match values that start with the provided prefix. */
   startsWith?: string;
+} | {
   /** Match values against an RE2 regular expression. */
   regex?: string;
 };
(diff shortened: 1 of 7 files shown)


packages/vercel-sandbox/src/network-policy.ts
```diff
@@ -11,10 +11,54 @@ export type NetworkTransformer = {
...
+ * Defines how a request value is matched. Exactly one field should be set.
+ */
+export type NetworkPolicyMatcher = {
+  /** Match the value exactly. */
+  exact?: string;
+  /** Match values that start with the provided prefix. */
+  startsWith?: string;
+  /** Match values against an RE2 regular expression. */
+  regex?: string;
+};
```
tutman96, line 26: Would it make sense to do a type union here since only one of these can be set?
```suggestion
export type NetworkPolicyMatcher = {
  /** Match the value exactly. */
  exact: string;
} | {
  /** Match values that start with the provided prefix. */
  startsWith: string;
} | {
  /** Match values against an RE2 regular expression. */
  regex: string;
};
```

tutman96 (approved):

QuiiBz:
https://github.com/vercel/sandbox/pull/173

PR #184, opened 2026-05-11
feat(dynamic ports): support dynamic update of sandbox ports

valerian-roche:
Users are currently required to provide the exact list of ports they will use when creating their sandbox. This can be constraining, and we received requests to support dynamically adding or removing ports on a running sandbox (akin to the network policy model)
This PR adds support for the latest `sandbox update` API extension for ports. Users can provide a new list of ports to expose (as a full state). New ports will become exposed and ones no longer presents will be deregistered from the sandbox.
The port number limit (currently 15) remains applicable

### DIFF
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -1154,22 +1154,6 @@ export class Sandbox {
   }
 
   /**
-   * Replace the exposed ports for this sandbox.
-   *
-   * This is a full-state update: any currently exposed port omitted from
-   * `ports` will be deregistered.
-   *
-   * @param ports - Complete list of ports to expose.
-   * @param opts - Optional abort signal.
-   */
-  async updatePorts(
-    ports: number[],
-    opts?: { signal?: AbortSignal },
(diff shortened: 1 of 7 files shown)


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -1153,9 +1153,28 @@ export class Sandbox {
...
+  /**
+   * Replace the exposed ports for this sandbox.
+   *
+   * This is a full-state update: any currently exposed port omitted from
+   * `ports` will be deregistered.
+   *
+   * @param ports - Complete list of ports to expose.
+   * @param opts - Optional abort signal.
+   */
+  async updatePorts(
```
QuiiBz: We deprecated `updateNetworkPolicy` in favor of `update`, so I don't think we should add a new `updatePorts` method here

QuiiBz (approved):

packages/vercel-sandbox/src/sandbox.ts
valerian-roche: Removed `updatePorts`

PR #188, opened 2026-05-14
feat(sdk): add `defineSandboxProxy` helper

QuiiBz:
This PR adds a new `defineSandboxProxy` helper in `@vercel/sandbox/proxy` to integrate [network policies forwarding rules](https://vercel.com/changelog/vercel-sandbox-firewall-now-supports-request-proxying-and-filtering) easily within a Vercel Function, or any other platform that supports the Web Handlers syntax (`Request`/`Response` objects)

We use `jose` to verify the OIDC token, then extract the metadata (original host, scheme, port, path, and source team/project/sandbox ids) before calling a user-defined Web Handler. If the request cannot be authorized, we return a 403 by default, but the user can also override this behavior with a separate Web Handler. A new `Request` is constructed to correspond to the original sandbox request before being proxied by the sandbox firewall

Example usage in a Vercel Function:

```ts
export default {
  fetch: defineSandboxProxy(async (request, meta) => {
    // meta contains the original host & source team/project/sandbox ids
    console.log(meta)

    // return a custom response, or proxy upstream:
    return await fetch(request)
  }, (request, error) => {
    // optional, handle any authorization error
    return new Response("Forbidden", { status: 403 })
  })
}
```

### DIFF
--- a/packages/vercel-sandbox/src/proxy.ts
+++ b/packages/vercel-sandbox/src/proxy.ts
@@ -58,6 +58,20 @@ export type InvalidRequestProxyHandler = (
  * 
  * This function automatically verifies the OIDC token included in proxied requests to ensure the request
  * is legitimate: if invalid, the `invalidRequestHandler` is called (by default, a 403 response is returned).
+ *
+ * @example
+ * ```ts
+ * export default {
+ *   fetch: defineSandboxProxy(async (request, meta) => {
+ *     console.log(meta) // contains the original host/scheme/port & source team/project/sandbox ids
+ *
+ *     return new Response("Response from my proxy");
+ *   }, (request, error) => {
+ *     // optional, handle any authorization error
+ *     return new Response("Forbidden", { status: 403 });
+ *   })
+ * }
+ * ```
  * 
  * @see https://vercel.com/docs/vercel-sandbox/concepts/firewall#requests-proxying
  */
--- a/packages/vercel-sandbox/src/proxy.ts
+++ b/packages/vercel-sandbox/src/proxy.ts
@@ -63,12 +63,19 @@ export type InvalidRequestProxyHandler = (
  * ```ts
  * export default {
  *   fetch: defineSandboxProxy(async (request, meta) => {
- *     console.log(meta) // contains the original host/scheme/port & source team/project/sandbox ids
+ *     // meta contains the original host/scheme/port & source team/project/sandbox ids
+ *     console.log(meta)
  *
- *     return new Response("Response from my proxy");
+ *     // return a custom response, or proxy upstream:
+ *     const url = new URL(request.url)
+ *     return await fetch(`${meta.scheme}://${meta.host}:${meta.port}/${url.pathname}${url.search}`, {
+ *       method: request.method,
+ *       headers: request.headers,
+ *       body: request.body
+ *     })
  *   }, (request, error) => {
  *     // optional, handle any authorization error
- *     return new Response("Forbidden", { status: 403 });
+ *     return new Response("Forbidden", { status: 403 })
  *   })
  * }
  * ```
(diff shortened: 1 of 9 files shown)


ctgowrie (approved):

packages/vercel-sandbox/src/proxy.ts
```diff
@@ -0,0 +1,220 @@
...
+      );
+    } catch {
+      return invalidRequestHandler(
+        new Request(request, { headers }),
+        new Error("Invalid proxied request URL"),
+      );
+    }
+
+    sanitizedRequest.headers.set("host", host);
+    sanitizedRequest.headers.set("x-forwarded-host", host);
```
valerian-roche: Can you clarify why we set this value here?

packages/vercel-sandbox/src/proxy.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  if (
+    issuerUrl.protocol !== "https:" ||
+    issuerUrl.hostname !== VERCEL_OIDC_HOSTNAME
+  ) {
+    throw new Error("Invalid OIDC issuer");
+  }
+
+  const { payload } = await jwtVerify(token, getJwks(issuer), {
+    audience: originalUrl.origin + originalUrl.pathname,
```
valerian-roche: The path of the request is "proxy URL + original path", so this is likely to not match here, no?

QuiiBz: Hum I only tested the `/` case, for example:
- Rule on `vercel.com` with `forwardURL: "https://proxy.vercel.app/proxy"`, which is the audience of the OIDC token
- I make a request to `https://vercel.com` from the sandbox
- The function at `https://proxy.vercel.app/proxy` receives the request, and has an `originalUrl` of `https://proxy.vercel.app/proxy` & `vercel-forwarded-host` of `vercel.com`

This indeed doesn't handle requests on other paths like `https://vercel.com/sandbox`, because now the audience would still be `https://proxy.vercel.app/proxy` but `originalUrl` would be `https://proxy.vercel.app/proxy/sandbox`

We could check only the hostname but that defeats the point of having path-defined audiences. The other option is to explicitly require the user to pass in their configured `forwardURL` into `defineSandboxProxy`. Thoughts?

packages/vercel-sandbox/src/proxy.ts
QuiiBz: Removed `x-forwarded-host` and only kept the overwrote of `host`. If you proxy the request as-is without overriding `host`, we'd have a mismatch with the SNI because it'd point to the proxy host

packages/vercel-sandbox/src/proxy.ts
QuiiBz: Following our discussion, updated to extract `vercel-forwarded-path` and derive the correct audience by removing it from the request URL

packages/vercel-sandbox/src/proxy.ts
```diff
@@ -0,0 +1,260 @@
...
+) {
+  return async function sandboxProxy(request: Request): Promise<Response> {
+    const headers = new Headers(request.headers);
+    const host = headers.get(FORWARDED_HOST_HEADER);
+    const scheme = headers.get(FORWARDED_SCHEME_HEADER);
+    const port = headers.get(FORWARDED_PORT_HEADER);
+    const path = headers.get(FORWARDED_PATH_HEADER);
+    const oidcToken = headers.get(SANDBOX_OIDC_TOKEN_HEADER);
+
+    headers.delete(FORWARDED_HOST_HEADER);
```
valerian-roche, line 85: To be more robust on header addition later should we drop all `vercel-forwarded-*` headers?

valerian-roche (approved):

packages/vercel-sandbox/src/proxy.ts
QuiiBz: :maybe: for now I don't think we need to, but we can always iterate

PR #196, opened 2026-05-19
feat(sdk,cli): add support for drives

QuiiBz:
This PR adds support for drives in beta. Drives are persistent storage that can be attached or detached to sandboxes. You can mount up to 4 drives per sandbox, with up to 1 TiB per drive

CLI:
```bash
sbx drives ls
sbx drives get-or-create --max-size <bytes>
sbx drives delete <drive>
sbx create --mount <drive>:/path[:read-only|read-write]
```

SDK:
```ts
const drive = await Drive.getOrCreate({
  name:  'my-drive',                           // unique per project
  maxSize: 1024 * 1024 * 1024 * 100  // optional, default 100 GiB
})

  name: 'my-sandbox',
  mounts: {
    '/mnt/storage': {               // can add up to 4 mounts
      drive: drive.name,
      mode: 'read-only'          // optional, defaults to 'read-write'
    }
  }
})

const { drives, pagination } = await Drive.list({ limit, since, until })
await drives.delete()
```


packages/sandbox/docs/index.md
```diff
@@ -341,6 +344,22 @@ Commands:
...
+
+▲ sandbox volumes [options] <command>
+
+For command help, run `sandbox volumes <command> --help`
+
+Commands:
+
+    ls | list                        List volumes for the specified account and project.
+    get-or-create  <name>            Create a volume if it does not already exist, or retrieve it.
+    rm | delete    <name> [...name]  Delete one or more volumes.
```
marc-vercel: Shouldn't this be `rm | remove` or `del | delete`? The abbreviation `rm` is not from the same word as `delete`.

packages/sandbox/src/commands/drives.ts
```diff
@@ -0,0 +1,228 @@
...
+  token: string;
+  teamId: string;
+  projectId: string;
+  name: string;
+}): Promise<Volume> {
+  const { volumes } = await volumeClient.list({
+    token,
+    teamId,
+    projectId,
+    namePrefix: name,
```
marc-vercel, line 211: Cannot we do a limit 1 here? I understand that the exact name will always be the first result, when sorted `desc` (here we are sorting `asc`). I might be wrong tho.

packages/vercel-sandbox/src/drive.ts
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Serialize a Volume instance to plain data for @workflow/serde.
+   *
+   * @param instance - The Volume instance to serialize
+   * @returns A plain object containing volume metadata
+   */
+  static [WORKFLOW_SERIALIZE](instance: Volume): SerializedVolume {
+    return {
+      volume: instance.volume,
+      projectId: instance._projectId,
```
marc-vercel, line 114: Just to confirm, but is it ok to only serialize these two parameters? I understand we do not serialize the others (e.g. namedSandboxId) because they might become obsolete, but this might break client-side applications between steps.

This is the same as if you picked a volume, did a sleep of 2s and then continued working with the volume, you expect that in most cases the sessionId and namedSandboxId is not obsolete.

For example, with Sandbox we serialize all of it:

```
  static [WORKFLOW_SERIALIZE](instance: Sandbox): SerializedSandbox {
    return {
      metadata: instance.session?._sessionSnapshot!,
      routes: instance.session?.routes ?? [],
      sandboxMetadata: instance.sandbox,
      projectId: instance.projectId,
    };
  }
```

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -35,12 +37,10 @@ import { Readable } from "stream";
...
 import { getVercelOidcToken } from "@vercel/oidc";
 import { NetworkPolicy } from "../network-policy.js";
-import {
-  toAPINetworkPolicy,
-  fromAPINetworkPolicy,
-} from "../utils/network-policy.js";
+import { toAPINetworkPolicy } from "../utils/network-policy.js";
 import { getPrivateParams, WithPrivate } from "../utils/types.js";
 import { RUNTIMES } from "../constants.js";
+import { BaseCreateSandboxParams } from "../sandbox.js";
```
marc-vercel:
```suggestion
import { type BaseCreateSandboxParams } from "../sandbox.js";
```

To avoid a circular import (sandbox.js also imports api-client).

packages/vercel-sandbox/src/volume.ts
```diff
@@ -0,0 +1,254 @@
...
+   */
+  public get name(): string {
+    return this.volume.name;
+  }
+
+  /**
+   * The project ID that owns the volume.
+   */
+  public get projectId(): string {
+    return this.volume.projectId;
```
marc-vercel: Shouldn't we use here `_projectId`, as the value might differ?

```
this._projectId = projectId ?? volume.projectId;
```

marc-vercel (approved):

packages/sandbox/src/commands/drives.ts
QuiiBz: Updated to limit 1 and removed the `find`; we need to keep `asc` in this case

packages/sandbox/docs/index.md
QuiiBz: We do support `rm`, `remove` and `delete`, updated that doc for both snapshots and volumes to `rm | remove` to match `sandbox rm|remove`

packages/sandbox/src/commands/drives.ts
QuiiBz: Actually reverting because Vercel agent added some good comments https://github.com/vercel/sandbox/pull/196#discussion_r3279813975 https://github.com/vercel/sandbox/pull/196#discussion_r3279813980
I'd prefer to be sure we're deleting the right volume

packages/vercel-sandbox/src/drive.ts
QuiiBz: The `Volume` object contains all the metadata we need, it's serializing everything

PR #203, opened 2026-05-22
feat(cli): allow updating sandbox ports

QuiiBz:
In https://github.com/vercel/sandbox/pull/184 we added support for updating a sandbox ports at runtime. However, we missed adding this functionnality to the CLI

This PR adds a new `sandbox config ports` subcommand to update the ports dynamically:
```sh
sandbox config ports my-sandbox-name # clear all ports
sandbox config ports my-sandbox-name -p 8080 -p 1234 # replaces currently opened ports
```
Screenshot 2026-05-22 at 08 59 01

We also surface the configured ports in `sandbox config list`:
Screenshot 2026-05-22 at 08 53 51

### DIFF
--- a/packages/sandbox/src/args/ports.ts
+++ b/packages/sandbox/src/args/ports.ts
@@ -14,7 +14,7 @@ export const publishPorts = cmd.multioption({
             [
               `Invalid port: ${number}.`,
               `${chalk.bold("hint:")} Ports must be integers between 1024-65535 (privileged ports 0-1023 are reserved).`,
-              "Examples: 3000, 8080, 8443",
+              "Examples: 3000, 8443",
             ].join("\n"),
           );
         }
--- a/packages/sandbox/src/args/ports.ts
+++ b/packages/sandbox/src/args/ports.ts
@@ -1,0 +1,25 @@
+import * as cmd from "cmd-ts";
+import chalk from "chalk";
+
+export const publishPorts = cmd.multioption({
+  long: "publish-port",
+  short: "p",
+  description: "Publish sandbox port(s) to DOMAIN.vercel.run",
+  type: cmd.array(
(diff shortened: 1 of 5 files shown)


packages/sandbox/src/args/ports.ts
```diff
@@ -0,0 +1,25 @@
...
+  type: cmd.array(
+    cmd.extendType(cmd.number, {
+      displayName: "PORT",
+      async from(number) {
+        if (!Number.isInteger(number) || number < 1024 || number > 65535) {
+          throw new Error(
+            [
+              `Invalid port: ${number}.`,
+              `${chalk.bold("hint:")} Ports must be integers between 1024-65535 (privileged ports 0-1023 are reserved).`,
+              "Examples: 3000, 8080, 8443",
```
marc-vercel: Let's remove 8080 for now until it is unblocked. WDYT?

packages/sandbox/src/commands/config.ts
```diff
@@ -658,6 +709,32 @@ function formatKeepLastSnapshots(
...
+
+function getPublishedRoutes(sandbox: Sandbox): Sandbox["routes"] | undefined {
+  try {
+    return sandbox.routes.filter(
+      (route) => route.port !== sandbox.interactivePort,
+    );
+  } catch (error) {
+    if (
+      error instanceof Error &&
+      error.message === "No active session. Run a command or call resume first."
```
marc-vercel: This will break if we change the message. Maybe it is better to check the error status (e.g. 4XX)?

marc-vercel (approved):
LGTM, just left some minor comments.

packages/sandbox/src/commands/config.ts
QuiiBz: We actually don't need this try catch, removed

PR #205, opened 2026-05-22
feat(vercel-sandbox,sandbox): add fork, remove --sandbox-snapshot

marc-vercel:
## Summary

Replaces the `--sandbox-snapshot` flag and `Snapshot.fromSandbox()` with a first-class `fork` primitive on both the SDK and the CLI:

- `Sandbox.fork({ sourceSandbox: "<name>", ... })` — copies the source sandbox's config (`resources`, `timeout`, `networkPolicy`, `tags`, `ports`, `persistent`, `snapshotExpiration`, `keepLastSnapshots`) and seeds the new sandbox from its current snapshot. Any field passed in `params` overrides the copied value. `env` is not copied (encrypted server-side) and must be re-supplied. If the source has no current snapshot, the fork falls back to creating a fresh sandbox with the source's runtime plus the copied config.
- `sandbox fork <source>` — CLI counterpart with the same override semantics.

Also refactors `commands/create.ts` to share `ports`, snapshot-retention args, the `buildKeepLastSnapshotsPayload` helper, and the `printSandboxSummary` printer with the new `fork` command.

CLI
---

A new top-level subcommand forks an existing sandbox by name. Without any flags, every supported config field is inherited from the source; any flag passed acts as an override. `env` is never inherited and must be re-supplied via `--env`.

```
> sandbox fork my-source
✅ Sandbox forked-sandbox-abc forked from my-source.
   │ team: my-team
   ╰ project: my-project

> sandbox fork my-source --name my-forked-sandbox --vcpus 4 --env FOO=1
✅ Sandbox my-forked-sandbox forked from my-source.
   │ team: my-team
   ╰ project: my-project
```

Override flags mirror `sandbox create`: `--name`, `--vcpus`, `--timeout`, `--publish-port`, `--env`, `--tag`, `--network-policy` (+ `--allowed-domain` / `--allowed-cidr` / `--denied-cidr`), `--snapshot-expiration`, `--keep-last-snapshots[-for]`, `--delete-evicted-snapshots`, `--non-persistent`, `--connect`, `--silent`. Tags fully replace the source's tags when provided (no per-key merge).

SDK
---

A new `Sandbox.fork` method exposes the same primitive:

```ts
// Inherit everything from the source
const fork = await Sandbox.fork({ sourceSandbox: "prod-agent" });

// Override specific fields; the rest are copied from the source
const fork = await Sandbox.fork({
  sourceSandbox: "prod-agent",
  name: "forked-prod-agent",
  resources: { vcpus: 4 },
  env: { OPENAI_API_KEY: process.env.OPENAI_API_KEY! },
});
```

Parameters:

- `sourceSandbox` (required): name of the sandbox to fork from.
- Any field accepted by `Sandbox.create` except `source` and `runtime`: when set, overrides the copied value; when omitted, the source's value is inherited.
- `env`: not copied (encrypted server-side); pass it explicitly to set environment variables on the fork.

When the source has a current snapshot, the fork is seeded from it (`source: { type: "snapshot", snapshotId }`). When it does not, the fork falls back to a fresh create with the source's `runtime` plus the copied config.

## Breaking

Removed `sandbox create --sandbox-snapshot <name>` and `Snapshot.fromSandbox(name)`. Use `sandbox fork <name>` / `Sandbox.fork({ sourceSandbox: name })` instead. Raw snapshot IDs are still supported via `--snapshot <id>` and `Sandbox.create({ source: { type: "snapshot", snapshotId } })`.


packages/sandbox/src/args/ports.ts
```diff
@@ -1,20 +1,20 @@
 import * as cmd from "cmd-ts";
```
marc-vercel, line 1: Moved from `create.ts`, as it is also shared by `fork.ts`

packages/sandbox/src/args/snapshot-retention.ts
```diff
@@ -0,0 +1,51 @@
+import * as cmd from "cmd-ts";
```
marc-vercel, line 1: Moved from `create.ts`, as it is also shared by `fork.ts`

packages/sandbox/src/util/keep-last-snapshots.ts
```diff
@@ -0,0 +1,60 @@
+import ms from "ms";
```
marc-vercel, line 1: Moved from `create.ts`, as it is also shared by `fork.ts`

packages/sandbox/src/util/print-sandbox-summary.ts
```diff
@@ -0,0 +1,59 @@
+import chalk from "chalk";
```
marc-vercel, line 1: Moved from `create.ts`, as it is also shared by `fork.ts`

packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -156,6 +156,29 @@ export type CreateSandboxParams =
...
+ * @inline
+ */
+export type ForkSandboxParams = Omit<
+  BaseCreateSandboxParams,
+  "source" | "runtime"
+> & {
+  /**
+   * Name of the source sandbox to fork from.
+   */
+  sourceSandbox: string;
```
marc-vercel, line 179: I had to move it from `source` to `sourceSandbox` so that it does not conflict with `source`.

AndyW22 (approved):

PR #215, opened 2026-06-01
fix(sandbox): make interactive connect resilient on stopped/resumed sandboxes

marc-vercel:
## Problem

`sandbox connect` (and interactive `sandbox exec`) could hang indefinitely on `Waiting for connection...`, or fail in a confusing way, when run against a stopped sandbox that has to be resumed. It worked reliably against an already-running sandbox, which is why it only showed up intermittently after a stop/resume.

Several independent issues combined to produce this:

1. **Real connection errors were swallowed.** Once the connection handshake landed, the abort signal that stops the "did the command exit early?" check was *also* used to filter errors from `attach()`. So any failure that happened after the handshake (for example, the resumed session not yet exposing a route for the interactive port) was silently discarded instead of surfaced.

2. **The spinner kept the process alive.** The progress spinner's teardown called `ora.clear()`, which only erases the current frame but leaves its render interval running. That timer keeps Node's event loop alive, so on any early teardown the CLI would sit forever on the spinner instead of exiting.

3. **Early server exits were opaque.** When the in-sandbox interactive server exited before connecting, the CLI showed a generic "may have timed out" hint with no detail.

4. **The in-sandbox server trusted a stale config.** `pty-tunnel-server` decided whether a server was already running purely from a leftover config file and a liveness check on its recorded PID. Across a snapshot/resume that config is restored from the snapshot while the original process is gone, so a coincidentally-reused PID made it connect to a dead socket and exit.

## Solution

- Stop funneling `attach()` through the connection-established abort filter, so genuine connection failures propagate instead of being swallowed.
- Always `stop()` the spinner on teardown (not just `clear()`), so a failure before the connection is established can no longer hang the process.
- Include the in-sandbox server's stderr in the error when it exits before connecting, so the real cause is visible.
- Have `pty-tunnel-server` health-check a server before reusing it, and remove any leftover config before spawning a new one, so a stale config restored from a snapshot can no longer cause a connection to a dead socket.

Together these turn the previous silent hang into either a working connection or a fast, legible error.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -938,6 +938,10 @@ export class Sandbox {
       resume: true,
       signal,
     });
+    // Do not update the session if it was not resumed.
+    if (!response.json.resumed && this.session) {
+      return;
+    }
     this.session = new Session({
       client,
       routes: response.json.routes,
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -938,10 +938,6 @@ export class Sandbox {
       resume: true,
       signal,
     });
-    // Do not update the session if it was not resumed.
-    if (!response.json.resumed && this.session) {
-      return;
-    }
     this.session = new Session({
       client,
       routes: response.json.routes,
--- a/.changeset/fix-interactive-connect-resume.md
+++ b/.changeset/fix-interactive-connect-resume.md
@@ -1,0 +1,5 @@
+---
+"sandbox": patch
+---
+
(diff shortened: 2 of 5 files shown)


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -938,6 +938,10 @@ export class Sandbox {
       resume: true,
       signal,
     });
+    // Do not update the session if it was not resumed.
+    if (!response.json.resumed && this.session) {
```
QuiiBz: Why do we need to do this now?

QuiiBz (approved):

PR #222, opened 2026-06-04
feat(sandbox,@vercel/sandbox): remove interactive-shell binary

marc-vercel:
Remove the interactive-shell binary from the CLI/SDK, and just rely on an API endpoint (as of now experimental) to load the binary used to create the interactive shell.

This should improve:
- Speed when connecting to the sandbox.
- Other code-bases can open the shell, as now you only need to call an API endpoint and open a websocket connection.


Schniz (approved):
Image

great thing to move it off here. will be interesting to see the perf benefits 👀

PR #251, opened 2026-07-15
feat(sdk): clarify usage of `transform` and `forwardURL`

QuiiBz:
Within network policies, `transform` and `forwardURL` are exclusive and should not be used together

The API currently accepts this combinaison and doesn't reject it: `forwardURL` takes precedence and `transform` doesn't apply. We cannot currently update the API without a breaking change/new version because some customers might already accidentaly depend on it

This PR updates the JS doc and validation of the SDK only, so that customers upgrading it have a chance to remove any usage of both `transform` and `forwardURL` within the same network policy

### DIFF
--- a/packages/vercel-sandbox/src/api-client/validators.ts
+++ b/packages/vercel-sandbox/src/api-client/validators.ts
@@ -39,11 +39,17 @@ export const NetworkPolicyTransformValidator = z.object({
   headers: z.record(z.string(), z.string()).optional(),
 });
 
-export const NetworkPolicyRuleValidator = z.object({
-  match: RuleMatchValidator.optional(),
-  transform: z.array(NetworkPolicyTransformValidator).optional(),
-  forwardURL: z.string().optional(),
-});
+export const NetworkPolicyRuleValidator = z
+  .object({
+    match: RuleMatchValidator.optional(),
+    transform: z.array(NetworkPolicyTransformValidator).optional(),
+    forwardURL: z.string().optional(),
+  })
+  .refine(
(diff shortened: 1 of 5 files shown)


packages/vercel-sandbox/src/api-client/validators.ts
```diff
@@ -39,11 +39,22 @@ export const NetworkPolicyTransformValidator = z.object({
...
+export const NetworkPolicyRuleValidator = z
+  .object({
+    match: RuleMatchValidator.optional(),
+    transform: z.array(NetworkPolicyTransformValidator).optional(),
+    forwardURL: z.string().optional(),
+  })
+  .refine(
+    ({ transform, forwardURL }) =>
+      transform === undefined || forwardURL === undefined,
+    { message: "transform and forwardURL cannot be used together" },
```
marc-vercel, line 51:
```suggestion
      transform === undefined && forwardURL === undefined,
    { message: "transform and forwardURL cannot be used together" },
```

Also, the error messages shouldn't be `transform or forwardURL must be provided`? And the other `.refine` use `transform and forwardURL cannot be used together`

marc-vercel: The tests work, so nvm

marc-vercel (approved):

packages/vercel-sandbox/src/api-client/validators.ts
QuiiBz: Yeah it's slightly confusing, refine is enforcing the callback so the error message is only thrown when the callback returns `false`

PR #285, opened 2026-08-10
feat(sdk,cli): add support for `Drive`

QuiiBz:
Add support for [Drives](https://vercel.com/docs/sandbox/concepts/drives), persistent storage that can store TBs of data and be mounted on sandboxes at a specific path. Drives support point-in-time, read-only snapshots.

Get started by creating a drive, then mount it on a sandbox:

```ts
import { Sandbox, Drive } from '@vercel/sandbox';
 
const drive = await Drive.getOrCreate({ name: 'workspace-cache' });
 
// Mount as read-write once
  name: 'my-sandbox',
  mounts: {
    '/data': drive
  },
});
```

Create a snapshot to mount the same drive on multiple sandboxes at a time:

```ts
import { Sandbox, Drive } from '@vercel/sandbox';
 
const drive = await Drive.getOrCreate({ name: 'shared' });
 
// Mount many snapshots of the same drive
const firstReader = await Sandbox.create({
  name: 'reader-1',
  mounts: {
    '/data': drive.snapshot()
  },
});
const secondReader = await Sandbox.create({
  name: 'reader-2',
  mounts: {
    '/data': drive.snapshot()
  },
});
```

Read the documentation to learn more: https://vercel.com/docs/sandbox/concepts/drives


packages/sandbox/docs/index.md
```diff
@@ -1,7 +1,7 @@
 ## `sandbox --help`
 
 ```
-sandbox 4.3.0
+sandbox 4.4.0-beta.0
```
marc-vercel: I think we should ignore the changes from `packages/sandbox/docs/index.md`. Changeset will update it automatically when creating its own PR.

packages/sandbox/src/commands/drives.ts
```diff
@@ -0,0 +1,244 @@
...
+  console.log(
+    table({
+      rows: drives,
+      columns: {
+        NAME: { value: (v) => v.name },
+        REGION: { value: (v) => v.region },
+        CREATED: { value: (v) => timeAgo(v.createdAt) },
+        UPDATED: { value: (v) => timeAgo(v.updatedAt) },
+        SIZE: { value: formatDriveSize },
+        ["ATTACHED SANDBOX"]: { value: (v) => v.currentSandboxName ?? "-" },
```
marc-vercel, line 201: Now drives can be attached as read-only to multiple sandboxes no? Or each drive snapshot = different drive ID?

packages/sandbox/src/commands/drives.ts
```diff
@@ -0,0 +1,244 @@
...
+        ["ATTACHED SESSION"]: { value: (v) => v.currentSessionId ?? "-" },
+      },
+    }),
+  );
+}
+
+function formatDriveSize(drive: Drive): string {
+  return drive.maxSize === undefined ? "-" : formatBytes(drive.maxSize);
```
marc-vercel: Shouldn't we show the default of 1 TiB here, instead of `-`?

packages/sandbox/src/commands/drives.ts
```diff
@@ -0,0 +1,244 @@
...
+  description: "Create a drive if it does not already exist, or retrieve it.",
+  args: {
+    name: cmd.positional({
+      type: driveName,
+      description: "Drive name to create or retrieve",
+    }),
+    maxSize: cmd.option({
+      long: "max-size",
+      description:
+        "Maximum drive size in bytes. If omitted, a default of 100 GiB is used.",
```
marc-vercel: Now it is 1 TiB no?

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -638,6 +643,59 @@ export class APIClient extends BaseClient {
     );
   }
 
+  async listDrives(params: {
+    projectId: string;
+    limit?: number;
+    cursor?: string | number;
+    since?: number | string;
+    until?: number | string;
```
marc-vercel: Is this supported? Afaik we paginate by `cursor` no? `since` and `until` are the old ways of paginating.

packages/sandbox/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sandbox",
   "description": "Command line interface for Vercel Sandbox",
-  "version": "4.3.0",
+  "version": "4.4.0-beta.0",
```
marc-vercel: Here (and in the other packages) we are leaking the `beta` into the main branch. This should be fixed when merging https://github.com/vercel/sandbox/pull/333 into this branch.

marc-vercel (approved):

packages/sandbox/package.json
QuiiBz: Yep that's the goal is that PR

packages/sandbox/src/commands/drives.ts
QuiiBz: That can actually be cleaned up, we always have a `maxSize` non-undefined value

packages/sandbox/src/commands/drives.ts
QuiiBz: Missed updating that one, good catch

packages/sandbox/src/commands/drives.ts
QuiiBz: Same drive id for snapshots, but only drives mounted as read-write will show up here

AndyW22 (approved):

PR #295, opened 2026-08-12
fix(cli): remove first-run dead ends (connect hint, staleness notice, ls hints, sh steer)

erulkey:
# PR: fix(cli): remove first-run dead ends

Branch: `cli-onboarding-dead-ends` · repo: vercel/sandbox · changeset: patch

## What

Four small DX fixes so CLI output never ends at a dead end:

1. **`create`/`fork` end with a connect hint** — `╰ connect with: sandbox ssh <name>`.
   Composing commands (`run`) suppress it via an internal `__printConnectHint` flag,
   and `--connect`/`sh` don't print it (you're already inside).
2. **Staleness notice** — `create`/`sh`/`run` fetch the latest published version from
   the npm registry concurrently with the creation request (zero added latency,
   1.5s time-boxed, failures silent) and print an update notice after the summary
   when the install is behind. Prereleases opt out. Disable with
   `SANDBOX_SKIP_VERSION_CHECK=1`.
3. **Sandbox-lookup 404s suggest `sandbox ls`** — added to the API-error formatter
   when the failing URL is a `/sandboxes/` route.
4. **`sandbox sh <command>` explains itself** — instead of "Unknown arguments", it
   says `sh` takes no command and points at `sandbox run --rm -i <command>`.

## Why

New-user friction audit (Aug 10-12): a stale globally-installed CLI silently boots
the pre-VMI image with no coding agents (the default is chosen client-side); `create`
output gave no next step, and users intuitively typed `sh` — landing in their LOCAL
shell; `connect` typos dumped a raw API URL with no recovery path; and
`sandbox sh claude` is the most natural post-launch command to type.

## Verification

- `pnpm typecheck` clean; 19/19 unit tests pass (new: isOutdated semver compare,
  404-hint present for /sandboxes/ URLs and absent otherwise).
- Live-run against production API with the local build:
  - `create --name hint-test` → summary ends `╰ connect with: sandbox ssh hint-test`
  - `run --rm -- echo run-ok` → hint suppressed, output unchanged otherwise
  - `sh claude` → steer message with `run --rm -i claude`
  - `connect doesnotexist-xyz` → 404 error now includes the `sandbox ls` hint
- NOT_TESTED: staleness notice end-to-end (local 4.0.0 == npm latest, so nothing to
  be behind); logic covered by unit tests.
- `pnpm generate-docs` re-run; docs/index.md included.

Follow-up candidates (out of scope): `connect` with no args attaching to the most
recently created sandbox; quieter 404 output with details behind a verbose flag.


packages/sandbox/docs/index.md
```diff
@@ -11,7 +11,7 @@ Commands:
 
     ls | list                                  List all sandboxes for the specified account and project.
     create                                     Create a sandbox in the specified account and project.
-    sh                                         Create a sandbox and start an interactive shell
+    sh             [...command]                Create a sandbox and start an interactive shell
```
Schniz, line 14: not sure that this is what we wanted

erulkey:
Split per review into a stack of four, one changelog entry each:

- #296 [sandbox-cli] end create and fork output with a connect hint
- #297 [sandbox-cli] print an update notice when the CLI is stale (stacked on #296)
- #298 [sandbox-cli] suggest sandbox ls on sandbox-lookup 404s (independent)
- #299 [sandbox-cli] steer sandbox sh <command> to run --rm -i (independent)

Together they reproduce this PR's diff exactly (create.ts is the only shared file, split between #296/#297). Keeping this open until the stack is reviewed, then it closes superseded.

erulkey:
Superseded — split per change and all four are now merged: #296 (connect hint), #297 (staleness notice), #298 (ls hint on 404), #299 (sh steer).

PR #298, opened 2026-08-13
[sandbox-cli] quiet lookup 404s with a sandbox ls hint

erulkey:
Third of four PRs splitting #295 per change (independent, no stack).

## What

A 404 on a `/sandboxes/` route becomes a quiet, actionable error: the not-found message, the status code, and a closing `run sandbox ls` hint. The requested-url and response-buffer lines move behind `DEBUG=sandbox:errors` (the buffer file is only written in debug mode).

Default:

```
Named sandbox 'my-sandbox' not found for this project.
├▶ status code: 404 Not Found
╰▶ hint: run sandbox ls to see the sandboxes in this project.
```

With `DEBUG=sandbox:errors`, the full form (url + buffer path) is restored, with the hint included.

## Why

New-user friction audit: `sandbox connect <unknown-name>` dumps the raw API URL (teamId, internal query params, a .http buffer path) with no next step. Debugging gold, onboarding poison — the hint gives the user their way forward.

## Changelog entry (changeset included)

> Sandbox-lookup 404 errors now suggest running `sandbox ls` to see your sandboxes instead of ending at a raw API error.

## Verification

- format-error unit tests: 17/17 pass (quiet default + DEBUG restore covered)
- `pnpm typecheck` clean

### DIFF
--- a/packages/sandbox/src/util/format-error.ts
+++ b/packages/sandbox/src/util/format-error.ts
@@ -38,7 +38,7 @@ export async function formatApiError(
       [
         message,
         `├▶ status code: ${status} ${error.response.statusText}`,
-        `╰▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls")} to see the sandboxes in this project.`,
+        `╰▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls --all")} to see the sandboxes in this project.`,
       ].join("\n"),
       error,
     );
@@ -52,7 +52,7 @@ export async function formatApiError(
   ];
   if (isLookupMiss) {
     lines.push(
-      `├▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls")} to see the sandboxes in this project.`,
+      `├▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls --all")} to see the sandboxes in this project.`,
     );
   }
   lines.push(
--- a/packages/sandbox/src/util/format-error.ts
+++ b/packages/sandbox/src/util/format-error.ts
@@ -30,8 +30,15 @@ export async function formatApiError(
     message,
     `├▶ requested url: ${error.response.url}`,
     `├▶ status code: ${status} ${error.response.statusText}`,
-    `╰▶ ${chalk.bold("hint:")} the full response buffer is stored in ${chalk.italic(tmpPath)}`,
   ];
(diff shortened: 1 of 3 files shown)


packages/sandbox/src/util/format-error.ts
```diff
@@ -22,16 +22,42 @@ const ApiErrorResponse = z.object({
...
+  const tmpPath = await writeResponseToTemp(error);
   const lines = [
     message,
     `├▶ requested url: ${error.response.url}`,
     `├▶ status code: ${status} ${error.response.statusText}`,
-    `╰▶ ${chalk.bold("hint:")} the full response buffer is stored in ${chalk.italic(tmpPath)}`,
   ];
+  if (isLookupMiss) {
+    lines.push(
+      `├▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls")} to see the sandboxes in this project.`,
```
marc-vercel: Maybe it should be `sandbox ls --all`, because `sandbox ls` only shows running sandboxes.

packages/sandbox/src/util/format-error.ts
```diff
@@ -22,16 +22,42 @@ const ApiErrorResponse = z.object({
...
+  // detail and the response buffer.
+  const isLookupMiss =
+    status === 404 && /\/sandboxes\//.test(error.response.url);
+  if (isLookupMiss && !debug.enabled) {
+    debug(`requested url: ${error.response.url}`);
+    return new StyledError(
+      [
+        message,
+        `├▶ status code: ${status} ${error.response.statusText}`,
+        `╰▶ ${chalk.bold("hint:")} run ${chalk.cyan("sandbox ls")} to see the sandboxes in this project.`,
```
marc-vercel: The same as https://github.com/vercel/sandbox/pull/298/changes#r3781934598

marc-vercel (approved):
Approved, but left two comments.

packages/sandbox/src/util/format-error.ts
erulkey: yes agreed - fixed

erulkey: int sites say sandbox ls --all now (6f9e5a1)

packages/sandbox/src/util/format-error.ts
erulkey: done in 6f9e5a1

PR #299, opened 2026-08-13
[sandbox-cli] steer sandbox sh <command> to run --rm -i

erulkey:
Fourth of four PRs splitting #295 per change (independent, no stack).

## What

`sandbox sh <command>` explains that `sh` takes no command and points at `sandbox run -i <command>`. Kept deliberately terse: no `--rm` in the suggested command (cleanup stays the user's choice, documented in run's own help) and no persistence footnote.

Revised per review (#295 discussion): the first version declared a `restPositionals` catch-all on `sh`, which made the generated help and docs advertise `sh [...command]` as if a command were accepted. The steer now runs as a pre-parse guard in the CLI entry on the leading non-flag tokens after `sh` (a leading token can never be an option value, so valid invocations like `sh --name my-box` can't be misread). `sh`'s argument surface, help, and generated docs are untouched; flag-tangled cases fall through to the parser's normal error.

## Why

`sandbox sh claude` is the most natural first thing to type after seeing the landing page, and it died with a generic "Unknown arguments". The user who types it wants exactly what the product can do; the error should say so.

## Changelog entry (changeset included)

> `sandbox sh <command>` now explains that `sh` takes no command and points at `sandbox run --rm -i <command>` instead of a generic "Unknown arguments" error.

## Verification

- steer-sh unit tests: 6/6 pass (steer, multi-token, plain sh, options-only, other commands, force-positional)
- `pnpm typecheck` clean
- Live from this branch build: `sandbox sh claude` prints the steer (exit 1); `sandbox sh --help` shows no command argument; `docs/index.md` byte-identical to main

### DIFF
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sandbox",
   "description": "Command line interface for Vercel Sandbox",
-  "version": "4.0.0",
+  "version": "3.0.0",
   "scripts": {
     "clean": "rm -rf node_modules dist",
     "sandbox": "ts-node ./src/sandbox.ts",
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sandbox",
   "description": "Command line interface for Vercel Sandbox",
-  "version": "3.0.0",
+  "version": "4.0.0",
   "scripts": {
     "clean": "rm -rf node_modules dist",
     "sandbox": "ts-node ./src/sandbox.ts",
--- a/packages/sandbox/src/util/steer-sh.ts
+++ b/packages/sandbox/src/util/steer-sh.ts
@@ -1,0 +1,38 @@
+import chalk from "chalk";
+import { StyledError } from "../error";
+
+/**
+ * `sandbox sh claude` is the most natural way to ask for "run this command in
+ * a sandbox", but `sh` takes no command. Declaring a catch-all positional on
+ * `sh` would advertise a command argument in its help and generated docs
+ * (review feedback on #295), so the steer runs here, before the parser's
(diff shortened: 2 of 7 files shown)


packages/sandbox/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sandbox",
   "description": "Command line interface for Vercel Sandbox",
-  "version": "4.0.0",
+  "version": "3.0.0",
```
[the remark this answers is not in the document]
marc-vercel: ☝️ True, we should remove this change.

packages/sandbox/src/util/steer-sh.ts
```diff
@@ -0,0 +1,38 @@
...
+ * invocation like `sandbox sh --name my-box`. Anything trickier falls through
+ * to the parser's normal error.
+ */
+export function steerShCommand(args: string[]): void {
+  if (args[0] !== "sh") {
+    return;
+  }
+  const command: string[] = [];
+  for (const arg of args.slice(1)) {
+    if (arg.startsWith("-")) {
```
marc-vercel: Mmh doesn't this also exclude from example strings like:

```
sandbox sh example-command -t 0
```

And it would just show as example `example-command`, instead of `example-command -t 0`?

I think that if the first argument starts with `-` we should break as we do now (as far as I see all parameters start either by - or --), but otherwise we should not break so that we catch the whole command.

marc-vercel (approved):
Approved with a comment.

packages/sandbox/src/util/steer-sh.ts
erulkey: fixed in 89b1645: once the first token after sh is a non-flag, the whole tail is captured, so sandbox sh example-command -t 0 suggests run -i example-command -t 0

PR #300, opened 2026-08-14
[sandbox-sdk] adopt the new snapshot defaults

marc-vercel:
The API changed its snapshot defaults (endpoints with major versions), but the SDK, CLI and Skill still targeted the previous versions.

These are the new defaults:

- Snapshots expire after **7 days** instead of 30.
- Persistent sandboxes get a `{ count: 1, deleteEvicted: true }` retention policy by default.
- `keepLastSnapshots: null` (CLI: `--keep-last-snapshots 0`) disables the limit and keeps every snapshot until it expires.

### DIFF
--- a/.changeset/snapshot-defaults.md
+++ b/.changeset/snapshot-defaults.md
@@ -1,0 +1,7 @@
+---
+"@vercel/sandbox": major
+"@vercel/sandbox-mock": major
+"sandbox": major
+---
+
+**Breaking:** adopt the new snapshot defaults. Snapshots now expire after 7 days by default instead of 30, and persistent sandboxes get a `{ count: 1, deleteEvicted: true }` retention policy unless one is given: only the most recent snapshot is kept, and taking a new one deletes the previous snapshot immediately instead of leaving it around until it expires. Pass `keepLastSnapshots: null` (CLI: `--keep-last-snapshots 0`) to keep every snapshot until it expires.
--- a/.changeset/snapshot-defaults.md
+++ b/.changeset/snapshot-defaults.md
@@ -4,4 +4,7 @@
 "sandbox": major
 ---
 
-**Breaking:** adopt the new snapshot defaults. Snapshots now expire after 7 days by default instead of 30, and persistent sandboxes get a `{ count: 1, deleteEvicted: true }` retention policy unless one is given: only the most recent snapshot is kept, and taking a new one deletes the previous snapshot immediately instead of leaving it around until it expires. Pass `keepLastSnapshots: null` (CLI: `--keep-last-snapshots 0`) to keep every snapshot until it expires.
+**\[Breaking\]** Reduced default snapshots retention:
+
+- Snapshots now expire after 7 days by default instead of 30 days previously. You can keep using expiration to set a custom snapshot expiration window in milliseconds, or set it to 0 to disable expiration.
+- Persistent sandboxes now preserve only the last session's snapshot by default, instead of the snapshot of each session. You can still set a custom retention policy via keepLastSnapshots.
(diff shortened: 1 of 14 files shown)


.changeset/snapshot-defaults.md
```diff
@@ -0,0 +1,7 @@
+---
+"@vercel/sandbox": major
+"@vercel/sandbox-mock": major
+"sandbox": major
+---
+
+**Breaking:** adopt the new snapshot defaults. Snapshots now expire after 7 days by default instead of 30, and persistent sandboxes get a `{ count: 1, deleteEvicted: true }` retention policy unless one is given: only the most recent snapshot is kept, and taking a new one deletes the previous snapshot immediately instead of leaving it around until it expires. Pass `keepLastSnapshots: null` (CLI: `--keep-last-snapshots 0`) to keep every snapshot until it expires.
```
QuiiBz: Something like this would be easier to understand:
> Reduced default snapshots retention:
> - snapshots now expire after 7 days by default instead of 30 days previously. You can keep using `expiration` to set a custom snapshot expiration window in milliseconds, or set it to 0 to disable expiration.
> - persistent sandboxes now preserve only the last session's snapshot by default, instead of the snapshot of each session. You can still set a custom retention policy via `keepLastSnapshots`.

skills/sandbox/SKILL.md
```diff
@@ -963,8 +965,9 @@ sandbox create --connect
 sandbox create --name my-app
 sandbox create --image my-repo:v1            # Boot from a VCR image
 sandbox create --non-persistent              # Disable filesystem persistence
-sandbox create --snapshot-expiration 7d      # Default snapshot TTL
-sandbox create --keep-last-snapshots 1       # Retention policy
+sandbox create --snapshot-expiration 7d      # Default snapshot TTL (7d when omitted)
+sandbox create --keep-last-snapshots 1       # Retention policy (persistent default: 1)
```
QuiiBz: So that it's different from the default:

```suggestion
sandbox create --snapshot-expiration 30d      # Default snapshot TTL (7d when omitted)
sandbox create --keep-last-snapshots 3       # Retention policy (persistent default: 1)
```

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -182,13 +182,13 @@ export class APIClient extends BaseClient {
...
         deleteEvicted?: boolean;
-      };
+      } | null;
       signal?: AbortSignal;
     }>,
   ) {
     const privateParams = getPrivateParams(params);
     const endpoint =
-      params.runtime === undefined ? "/v3/sandboxes" : "/v2/sandboxes";
+      params.runtime === undefined ? "/v4/sandboxes" : "/v2/sandboxes";
```
QuiiBz, line 191: Hum so if we keep using the deprecated `runtime`, we don't get the new defaults?

QuiiBz (approved):

.changeset/snapshot-defaults.md
marc-vercel: LGTM, less technical and more clear. Applied in [b3cc69c](https://github.com/vercel/sandbox/pull/300/commits/b3cc69c2dc1fc68cfacf3011d5a15c6c2cdfceb9) (I kept the **Breaking** header to make it more clear).

packages/vercel-sandbox/src/api-client/api-client.ts
marc-vercel: Hmm no, only customers that are using the latest endpoint version (contains `runtime` deprecation) can do that. This will also translate into customers using the next major SDK/CLI version.

PR #301, opened 2026-08-18
feat(sandbox,vercel-sandbox): multi-region support

marc-vercel:
Enable multi-region for sandboxes and snapshots. When we show the region, if it is not defined, we default to `iad1`, which is where we place sandboxes/snapshots by default.

Example of creating a sandbox (now we also show the region):

```
❯ sandbox create
✅ Sandbox silver-colossal-wren-WU4M8Y created.
   │ team: marc-codina-enhanced-vtest314
   │ region: iad1
   │ project: my-sandbox-app-tmp
   ╰ connect with: sandbox ssh silver-colossal-wren-WU4M8Y
```

Example of creating a sandbox with failover regions:

```
sandbox create --region cle1 --failover-regions sfo1,iad1
✅ Sandbox scarlet-vitreous-woodpecker-kkjYfK created.
   │ team: marc-codina-enhanced-vtest314
   │ region: cle1
   │ project: my-sandbox-app-tmp
   ╰ connect with: sandbox ssh scarlet-vitreous-woodpecker-kkjYfK
```

Example of updating an existing sandbox region and failover regions:

```
❯ sandbox config list crimson-chronic-mackerel-pbh3WM
FIELD                 VALUE                                 
Region                cle1                                  
Failover regions      sfo1, iad1                            
...

❯ sandbox config failover-regions crimson-chronic-mackerel-pbh3WM sfo1
✅ Configuration updated for sandbox crimson-chronic-mackerel-pbh3WM
   ╰ failover-regions: sfo1

❯ sandbox config region crimson-chronic-mackerel-pbh3WM iad1
✅ Configuration updated for sandbox crimson-chronic-mackerel-pbh3WM
   ╰ region: iad1

❯ sandbox config list crimson-chronic-mackerel-pbh3WM
FIELD                 VALUE    
Region                iad1     
Failover regions      sfo1
...
```

Notes:
- We are not validating the regions, so that if in the future we add more, old sandbox versions keep working.

PR #307, opened 2026-08-20
feat(sdk,cli): multi-region support for drives

QuiiBz:
Following https://github.com/vercel/sandbox/pull/301, this PR adds support for multi-region in drives

- Set the `region` field when creating a new drive, or use the `--region` flag in the CLI
- You can retrieve the region of a drive using the `region` getter on the `Drive` class
- Drives can be only be mounted to sandboxes running on the same region

Also adding the missing mock for the `Drive` class in `@vercel/sandbox-mock`


packages/vercel-sandbox/src/drive.ts
```diff
@@ -67,6 +73,13 @@ export class Drive {
     return this._projectId;
   }
 
+  /**
+   * The region where the drive data is stored.
+   */
+  public get region(): string {
+    return this.drive.region ?? DEFAULT_SANDBOX_REGION;
```
QuiiBz, line 80: Defaulting to `DEFAULT_SANDBOX_REGION` in case it was stored without it in a workflow step

AndyW22 (approved):

PR #312, opened 2026-08-21
fix(tags): allow only 1 tag on `Sandbox.list`

AndyW22:
We support listing with 1 tag via the API. This PR adds type safety so that only 1 tag can be used in the `tags` object when listing sandboxes. We can use a union to show a type error if more than 1 is passed.

Example:

image

### DIFF
--- a/packages/vercel-sandbox/src/sandbox.ts
+++ b/packages/vercel-sandbox/src/sandbox.ts
@@ -9,7 +9,10 @@ import { APIClient } from "./api-client/index.js";
 import { APIError } from "./api-client/api-error.js";
 import { type Credentials, getCredentials } from "./utils/get-credentials.js";
 import { getPrivateParams, type WithPrivate } from "./utils/types.js";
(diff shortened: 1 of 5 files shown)


packages/vercel-sandbox/src/sandbox.ts
```diff
@@ -621,9 +624,17 @@ export class Sandbox implements ExecutionContext {
    * // or: for await (const page of result.pages()) { ... }
    * ```
    */
-  static async list(
-    params?: Partial<Parameters<APIClient["listSandboxes"]>[0]> &
-      Partial<Credentials> &
+  static async list<Tags extends Record<string, string>>(
```
marc-vercel, line 627: This is still not validating the input from the user where we only allow one tag no? It just sends the request to the API and relies on the response. Maybe it is easier to validate in the CLI and return a proper message.

marc-vercel (approved):
Left a comment, LGTM

PR #319, opened 2026-08-27
[cli] add opt-out telemetry and AI-agent attribution

erulkey:
The Vercel CLI answers "which AI agents drive this tool" with `@vercel/detect-agent` plus its telemetry bridge; the Sandbox CLI has no telemetry at all, so agent-driven `sbx` usage is invisible. The only attribution we have today is the Vercel CLI's own `command:sandbox` events, and those already show the bundled path is majority agent-driven: of ~4.5k `vercel sandbox` sessions in the last 30 days, codex drove 2,346 and claude-code ~1,166, against 783 with no agent detected.

This:

- adds a telemetry client that posts generic key/value events (`subcommand`, `agent`, `version`, `platform`, `arch`, `ci`, `embedded`) to `https://telemetry.vercel.com/api/sandbox-cli/v1/events`, flushed once per invocation with a **1.5s** timeout and silent failure
- tags every API request's user-agent with the detected agent (`vercel/sandbox-cli/4.2.0 agent/claude-code_2-1-247_agent ...`), so the server side can attribute sandbox creation to agents once ingestion records the header
- adds `sandbox telemetry status|enable|disable`; `VERCEL_SANDBOX_TELEMETRY_DISABLED=1` or `VERCEL_TELEMETRY_DISABLED=1` also disable, `VERCEL_TELEMETRY_DEBUG=1` prints events to stderr without sending, and running as `vercel sandbox` respects the Vercel CLI's own global telemetry preference

Subcommand values are matched against an allowlist of registered command names, so an option value that precedes the subcommand (a token, a path) can never end up in an event. No flags, arguments, sandbox names, or file contents are recorded. I left per-sandbox identifiers out: the SDK exposes no public sandbox id, and the billing-grade agent-to-sandbox join belongs to the user-agent path on the server side anyway.

The `sandbox-cli` namespace is not registered on the bridge yet, so events are dropped server-side until companion changes land in vercel/data-schemas and vercel/public-telemetry-bridge (drafted, following the `sandbox_sdk_py` precedent). The CLI treats that as normal: a probe against the live endpoint returns `HTTP 400`, which flush swallows and debug-logs.

Verified:

- `pnpm vitest run`: 100 passed, 7 new in `src/telemetry/telemetry.test.ts`; `pnpm turbo typecheck --filter=sandbox` clean
- `VERCEL_TELEMETRY_DEBUG=1 pnpm sandbox ls` against production resolved the real team id and printed:

```
[telemetry] {"key":"subcommand","value":"ls","team_id":"team_6FYaZ0us0ywq0AkOcQMI0Rdy","session_id":"6561a21a-..."}
[telemetry] {"key":"agent","value":"claude-code_2-1-247_agent","team_id":"team_6FYaZ0us0ywq0AkOcQMI0Rdy","session_id":"6561a21a-..."}
```

- a local listener standing in for the bridge received the real POST with `client-id: sandbox-cli`, `x-sandbox-cli-topic-id: generic`, and a session header
- `telemetry disable` then `status` then `enable` round-trips through `~/Library/Preferences/sandbox-cli/telemetry.json`

An independent verification pass caught a stale receipt in the first version of this body: the typecheck claim predated the test file, and `tsc` rejected two untyped mock accesses. Fixed in 8d3f2e5; both `pnpm turbo typecheck --filter=sandbox --force` (4/4) and `pnpm vitest run` (100 passed) were re-run and observed green after that commit.

One known asymmetry I'm deliberately not addressing here: errors thrown during argument or auth resolution exit through cmd-ts's `Exit` effect (`process.exit()` before the `catch` in `sandbox.ts`), so those invocations never flush events. Errors from inside command handlers flush normally. Fixing it means moving to `runSafely` and owning cmd-ts's error printing; happy to do that as a follow-up if we want parse-error events.

The agent value comes from the `AI_AGENT` variable agents set themselves (Claude Code reports `claude-code_2-1-247_agent`), which is the same format the Vercel CLI's events carry, so the two datasets join on identical values.


alexfarnhambrown (commented):
Read the full diff plus the Vercel CLI side of the handoff. This closes a real blind spot — today `command='sandbox'` rows have a null subcommand on 234,010 of 234,483 invocations over 30 days.

Four things below. Two are bugs I could reproduce; two are fields that are free now and a schema version bump once the topic registers.

packages/sandbox/src/telemetry/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  const dir = xdgAppPaths("sandbox-cli").config();
+  mkdirSync(dir, { recursive: true });
+  writeFileSync(configFilePath(), JSON.stringify({ enabled }, null, 2));
+}
+
+/**
+ * Reads the Vercel CLI's global telemetry preference so that running as
+ * `vercel sandbox` respects an existing `vercel telemetry disable`.
+ */
+function vercelCliTelemetryDisabled(): boolean {
```
alexfarnhambrown, line 51: This path doesn't resolve: the Vercel CLI writes config.json to `XDGAppPaths('com.vercel.cli').dataDirs()[0]` (`packages/cli-config/src/paths.ts`), and this reads `.config()` — on my machine `~/Library/Preferences/com.vercel.cli` doesn't exist while `~/Library/Application Support/com.vercel.cli` holds `{"telemetry":{"enabled":true}}`. So `vercel telemetry disable` is silently not inherited, which makes the opt-out impossible to describe accurately in docs or in our coverage reporting. Recommend importing `getGlobalPathConfig()`/`getConfigFilePath()` from `@vercel/cli-config`, which also picks up the `~/.now` and legacy fallbacks.

packages/sandbox/src/client.ts
```diff
@@ -50,6 +64,13 @@ const fetchWithUserAgent: typeof globalThis.fetch = (input, init) => {
   );
   let agent = `vercel/sandbox-cli/${version}`;
 
+  // Attribute API traffic to the AI agent driving this invocation, if any,
+  // so the server side can record it once ingestion support lands.
+  const aiAgent = await detectAgentName();
```
alexfarnhambrown: This appends `agent/<name>` to every API request without checking whether telemetry is enabled, so opting out stops the event stream but still sends the detected agent on every request — which undercuts the opt-out story and leaves us describing two different consent states for one setting. Recommend gating it on the same `enabled` getter.

packages/sandbox/src/telemetry/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  updateTeamId(teamId: string | undefined): void {
+    if (teamId) this.teamId = teamId;
+  }
+
+  async trackInvocation(opts: { appName: string; argv: string[] }): Promise<void> {
+    this.embedded = opts.appName !== "sandbox";
+    // Only ever record known subcommand names, so option values that happen
+    // to precede the subcommand (tokens, paths) can never end up in an event.
+    const subcommand = opts.argv.find((arg) => SUBCOMMANDS.has(arg));
```
alexfarnhambrown: `argv.find(arg => SUBCOMMANDS.has(arg))` scans all of argv including tokens after `--`, so a registered command name arriving as an option value wins: `sandbox --scope create exec my-box` records `subcommand=create`. A silently wrong subcommand is worse for us than a null one because nothing flags it. Recommend stopping the scan at the first `--` and skipping known option-value pairs, or capturing what cmd-ts actually dispatched to. (Checked the allowlist against `app.ts` — complete today, including copy/cp, list/ls, connect/ssh/shell, remove/rm.)

packages/sandbox/src/telemetry/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  track(key: string, value: string | undefined): void {
+    if (!value) return;
+    this.events.push({ id: randomUUID(), event_time: Date.now(), key, value });
+  }
+
+  updateTeamId(teamId: string | undefined): void {
+    if (teamId) this.teamId = teamId;
+  }
+
+  async trackInvocation(opts: { appName: string; argv: string[] }): Promise<void> {
```
alexfarnhambrown, line 138: The description says the SDK exposes no public sandbox id, but `Sandbox.currentSession()` (`vercel-sandbox/src/sandbox.ts:1011`) and `Session.sessionId` (`session.ts:135`) are both public and return the `sbx_` id, which is the primary key of the table sandbox execution lands in (`product.analytics.sandbox_sessions`, populated on every row) — without it we can see a command ran but can't tell whether it produced a sandbox or what that sandbox did. Recommend emitting it on create/fork/exec/connect/stop, plus some marker for whether the invocation created a new sandbox or attached to an existing one, since sh/run/exec do both and adoption and re-engagement otherwise collapse into the same number.

packages/sandbox/src/index.ts
```diff
@@ -1,11 +1,17 @@
 import { run as runCmd } from "cmd-ts";
 import { app } from "./app";
+import { telemetry } from "./telemetry";
 
 export function createApp(opts: { withoutAuth: boolean; appName: string }) {
   const instance = app(opts);
   return {
     async run(args: string[]) {
-      await runCmd(instance, args);
+      await telemetry.trackInvocation({ appName: opts.appName, argv: args });
```
alexfarnhambrown, line 9: Nothing ties an event here back to the `vercel sandbox` invocation that spawned it — `embedded` proves it came through the Vercel CLI but names no invocation, and `sessionId` is a fresh randomUUID — so none of the user, team, project, device or CI context we already have on the Vercel CLI side can attach. Recommend forwarding the Vercel CLI's `invocation_id` from `packages/cli/src/commands/sandbox/index.ts` (currently forwards only scope/team/token) and emitting it as a key, plus `exit_code`, which is the only way we can tell whether a non-deploy command succeeded and is far cheaper to add here than to retrofit into the Vercel CLI.

packages/sandbox/package.json
```diff
@@ -55,6 +55,7 @@
     "chalk": "^5.6.0",
     "cmd-ts": "0.15.0",
     "date-fns": "^4.1.0",
+    "detect-agent": "1.2.0",
```
alexfarnhambrown: This pulls `detect-agent@1.2.0` while the Vercel CLI uses `@vercel/detect-agent`; they're different implementations — the monorepo one collapses `github-copilot-cli` → `github-copilot`, this one emits `AI_AGENT` verbatim, and their fallback ladders differ — so the two datasets join on identical values for most agents and diverge for some. We already have a ticket reconciling the CLI's vocabulary against the plugin's and a third vocabulary widens it. Recommend `@vercel/detect-agent` here.

packages/sandbox/src/telemetry/index.ts
alexfarnhambrown: Confirmed — `getConfigFilePath(getGlobalPathConfig())` is the right surface, that's the path the CLI actually writes to. Resolving.

packages/sandbox/src/client.ts
alexfarnhambrown: `telemetry.enabled` covers the config file and the embedded Vercel CLI opt-out as well as the env vars, so this is fully closed rather than just the env-var case. Resolving.

packages/sandbox/src/telemetry/index.ts
alexfarnhambrown: Anchoring on `argv[0]` with the allow-list solves it, and recording nothing on a leading flag is the better call over guessing. Resolving.

packages/sandbox/src/telemetry/index.ts
alexfarnhambrown: This is better than what I asked for. `sandbox_session_id` gives us the join, and `sandbox_session_origin` answers create-vs-attach, which I had down as a separate gap.

I checked the warehouse side: `product.analytics.sandbox_sessions.SANDBOX_SESSION_ID` is the same `sbx_` id, so this joins straight to shipped execution telemetry with no new modelling, and without depending on the `named_sandbox_id` work that's still in review. Resolving.

packages/sandbox/package.json
alexfarnhambrown: `@vercel/detect-agent@1.2.5` puts this on the same vocabulary as the Vercel CLI, which was the thing to avoid — a third agent naming scheme alongside CLI and plugin. Resolving.

packages/sandbox/src/index.ts
alexfarnhambrown: Following up — the receiving half is in, but I don't think anything sets the variable yet.

`flush()` reads `process.env.VERCEL_CLI_INVOCATION_ID` and #997 declares the matching nullable field, but I couldn't find anywhere that sets it: it isn't in `vercel/vercel` on main, and I didn't find an open PR there that adds it. If that's right, the field lands null on every event and embedded runs still can't be tied back to the `vercel sandbox` invocation that spawned them.

Looks like a small change on the CLI side — `packages/cli/src/commands/sandbox/index.ts` already has the value available via `TelemetryEventStore.currentInvocationId`, right next to where it sets `VERCEL_AUTH_TOKEN`. Different repo though, so it probably needs someone with CLI ownership rather than this PR.

One possible overlap worth flagging before anyone picks it up: vercel/vercel#16982 and #16990 look like they reimplement `vercel sandbox` natively and add `trackCliSubcommand*` for exec/create/connect/sh/fork/run. If those land, embedded subcommands would reach `cli_invocations` directly and this env-var plumbing might only be needed for the standalone CLI — or not at all. Both were last touched in early July so I don't know whether they're still live. Might be worth a quick word with Josh either way.

alexfarnhambrown (approved):
Approving from the data side. Agent vocabulary, opt-out path, subcommand anchoring, sbx session id + origin, and exit_code all resolved; the session id joins straight to product.analytics.sandbox_sessions.

The one open thread (VERCEL_CLI_INVOCATION_ID never being set) is a vercel/vercel change, not a blocker here — it can land null until that side is settled.

Sequencing note: the client swallows bridge responses, so data-schemas#997 and public-telemetry-bridge#37 should be live before a CLI release carrying this ships, otherwise events post to a 404 with no signal.

packages/sandbox/src/index.ts
alexfarnhambrown: Closing the loop: vercel/vercel#17600 sets `VERCEL_CLI_INVOCATION_ID` around the `createApp` call, so this lands non-null for embedded runs once both ship. Elisabeth is on it as reviewer.

One related ask now that decoupling is confirmed: standalone runs have no per-user identity at all, and the CLI never resolves the user after device-flow login. A `GET /v2/user` with the bearer token (what the Vercel CLI does for `whoami`) would give a `user_id` to emit alongside `team_id`; I've suggested the matching nullable field on data-schemas#997 so it's a code-only change here later. Not blocking this PR.

packages/sandbox/src/telemetry/index.ts
```diff
@@ -0,0 +1,205 @@
...
+  private teamId = "NO_TEAM_ID";
+  private projectId: string | undefined;
+  private embedded = false;
+
+  get enabled(): boolean {
+    if (
+      process.env.VERCEL_SANDBOX_TELEMETRY_DISABLED ||
+      process.env.VERCEL_TELEMETRY_DISABLED
+    ) {
+      return false;
```
marc-vercel, line 99: We have this same check in the other PR you created. After merging the first, can we ask the agent to move this to a generic function?

packages/sandbox/src/telemetry/agent.ts
```diff
@@ -0,0 +1,18 @@
...
+ * result is used both for telemetry events and for the user-agent header
+ * on every API request.
+ */
+export function detectAgentName(): Promise<string | undefined> {
+  if (!agentNamePromise) {
+    agentNamePromise = determineAgent().then(
+      (result) => (result.isAgent ? result.agent.name : undefined),
+      () => undefined,
+    );
+  }
```
marc-vercel, line 16: I think we have something similar in the other PR you created. After merging the first, can we ask the agent to move this to a generic function, if possible?

marc-vercel (approved):
LGTM, left some comments.

packages/sandbox/src/index.ts
```diff
@@ -1,11 +1,21 @@
...
+      try {
+        await runCmd(instance, args);
```
marc-vercel, line 11: Note: commands such as `exec` and batch operations report failures through `process.exitCode` without throwing.

Let's merge this as it is and we can fix that in another PR, as I do not think this will be a simple change.

packages/sandbox/src/index.ts
erulkey: the setter now exists: your vercel/vercel#17600 (approved) sets VERCEL_CLI_INVOCATION_ID around the bundled call, and this side already emits it on every event - null for standalone runs, which doubles as the channel discriminator. Resolving as closed by 17600

packages/sandbox/src/telemetry/index.ts
erulkey: agreed - once both PRs are in I'll move the shared detection + env-gate logic into @vercel/sandbox and have the CLI import it (follow-up PR)

packages/sandbox/src/telemetry/agent.ts
erulkey: same as the sibling comment: follow-up PR after both merge moves this into @vercel/sandbox as the single copy

packages/sandbox/src/index.ts
erulkey: good catch, agreed to handle in a follow-up: at flush time read process.exitCode when run() resolves (exec/batch failures then stop reporting 0). Merging as-is per your note

PR #320, opened 2026-08-27
[sdk] tag API requests with the detected AI agent

erulkey:
Stacked on #319, which added the `agent/<name>` user-agent phrase to the CLI. This extends the same stamp to the SDK, so programmatic callers get attributed too once the server side records the header (same ingestion ask, no new dependency on it).

This:

- appends ` agent/<name>` to the SDK's existing user-agent when `detect-agent` identifies an AI agent driving the process; no agent means the header is byte-identical to today
- adds the dedupe guard in the CLI's fetch wrapper so CLI-over-SDK traffic carries exactly one agent phrase, with a unit test pinning that
- adds `detect-agent@1.2.0` as an SDK runtime dependency (zero transitive dependencies, node >=20 which the SDK already requires)

Deliberately not added: telemetry events. The SDK ships in customer production code; this is request metadata only, nothing is collected or sent anywhere new.

Verified:

- probe with a logging fetch inside a Claude Code session:

```
UA: vercel/sandbox/3.2.0 agent/claude-code_2-1-247_agent (Node.js/v26.7.0; darwin/arm64)
```

- same probe with `env -u CLAUDECODE -u CLAUDE_CODE -u AI_AGENT` (a deployed server's environment): `vercel/sandbox/3.2.0 (Node.js/v26.7.0; darwin/arm64)`, unchanged
- combined path unit test asserts exactly one ` agent/` occurrence in the final header
- `pnpm turbo typecheck --filter=@vercel/sandbox --filter=sandbox --force`: 4/4; CLI suite 101 passed; SDK suite 291 passed, 80 skipped (pre-existing live-credential skips)

Detection only labels the process driving the SDK, so servers creating sandboxes to run agents inside them correctly carry no label; that population is the planned self-declared tag, not this PR.


packages/vercel-sandbox/package.json
```diff
@@ -56,6 +56,7 @@
     "@vercel/oidc": "3.2.0",
     "@workflow/serde": "4.1.0-beta.2",
     "async-retry": "1.3.3",
+    "detect-agent": "1.2.0",
```
alexfarnhambrown: This puts standalone `detect-agent` into the public SDK while the Vercel CLI uses `@vercel/detect-agent` — different implementations with different normalisation, so agent names won't always match across the streams we'd be joining. It's also the hardest of the three to change later, being a published dependency in customer code rather than a CLI. Recommend `@vercel/detect-agent`, or at least settling the canonical package before this ships.

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -125,12 +126,17 @@ export class APIClient extends BaseClient {
   protected async request(path: string, params?: RequestParams) {
     await this.ensureValidToken();
 
+    // Attribute traffic to the AI agent driving this process, if any, so the
+    // server side can record it once ingestion support lands.
+    const aiAgent = await detectAgentName();
```
alexfarnhambrown, line 131: The SDK has no telemetry setting, so once the server side records this header there's no way to opt out of agent attribution, and the CLI's `VERCEL_TELEMETRY_DISABLED` won't cover SDK traffic. That's request metadata now but collected data the moment ingestion lands, and cheaper to settle before a public SDK ships it. Recommend either honouring `VERCEL_TELEMETRY_DISABLED` here or documenting that this stamp isn't covered by it.

alexfarnhambrown (commented):
Checked where this header could land: no sandbox table in PRODUCT has a user_agent column — not the usage facts, not any sandbox event-streaming topic. So both this and #319 depend on an ingestion change that doesn't have a ticket or owner yet. If the landing place is the sandbox usage-fact pipeline, that's the same pipeline we need a failure status from, so it's probably one ask rather than two.

The upside is real though: the composed header separates CLI-driven from programmatic from agent-driven with no join required, which is a split we can't currently make.

packages/vercel-sandbox/package.json
alexfarnhambrown: Same fix as #319, so the SDK and the CLI now normalise agent names identically. Resolving.

packages/vercel-sandbox/src/api-client/api-client.ts
alexfarnhambrown: Half of this is closed — `detectAgentName()` now returns undefined when `VERCEL_TELEMETRY_DISABLED` or `VERCEL_SANDBOX_TELEMETRY_DISABLED` is set.

What's left is the config-file case: someone who ran `vercel telemetry disable` has `enabled: false` in config.json and no env var set, and the SDK can't see that, so they'd still get `agent/<name>` on the user-agent. The CLI path in #319 handles it through `telemetry.enabled`; the SDK has no equivalent. Looks like the automated review landed on the same point today.

Not urgent from our side — no sandbox table in PRODUCT has a `user_agent` column, so nothing reads this header yet. Leaving open in case you want to close the config-file case before something does.

rudivirawan69-cell (approved):

packages/vercel-sandbox/src/utils/detect-agent.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Detects the AI agent driving this process, if any. Memoized so the
+ * environment is inspected once per process, not once per request.
+ * Attribution honors the telemetry opt-out variables: disabling telemetry
+ * also stops the agent phrase from being added to the user-agent header.
+ */
+export function detectAgentName(): Promise<string | undefined> {
+  if (
+    process.env.VERCEL_TELEMETRY_DISABLED ||
+    process.env.VERCEL_SANDBOX_TELEMETRY_DISABLED
```
marc-vercel, line 14: I do not have much context here, but why are we using two environment variables to achieve the same thing?

I understand this is because we need to also be compatible with `vc` CLI? If so, cannot we all just use `VERCEL_TELEMETRY_DISABLED` here also?

packages/vercel-sandbox/src/api-client/api-client.ts
```diff
@@ -125,12 +126,17 @@ export class APIClient extends BaseClient {
...
+    const agentSuffix = aiAgent ? ` agent/${aiAgent}` : "";
+
     return super.request(path, {
       ...params,
       query: { teamId: this.teamId, ...params?.query },
       headers: {
         "content-type": "application/json",
-        "user-agent": `vercel/sandbox/${VERSION} (Node.js/${process.version}; ${os.platform()}/${os.arch()})`,
+        "user-agent": `vercel/sandbox/${VERSION}${agentSuffix} (Node.js/${process.version}; ${os.platform()}/${os.arch()})`,
```
marc-vercel, line 139: As far as I understand, here we are sending `vercel/sandbox/${VERSION}${agentSuffix}`, which will translate into:

`vercel/sandbox/v3.1.0agent/XXX (Node.js/YYY; linux/amd)`. Can we send the values like NodeJS and linux, so that we separate each part? For example:

```
`vercel/sandbox/v3.1.0 (agent/XXX; Node.js/YYY; linux/amd)`
```

packages/vercel-sandbox/src/utils/detect-agent.ts
marc-vercel: Oh I see, LGTM then: `VERCEL_SANDBOX_TELEMETRY_DISABLED` is so that we only target the sandbox.

marc-vercel (approved):
LGTM. Left a comment.

packages/vercel-sandbox/src/api-client/api-client.ts
erulkey: gated on VERCEL_TELEMETRY_DISABLED / VERCEL_SANDBOX_TELEMETRY_DISABLED (25a4b6c), and a config-file `sandbox telemetry disable` is enforced at the CLI layer by stripping the phrase from the header (8bcf7da, test-pinned). The docs page will state both

packages/vercel-sandbox/src/utils/detect-agent.ts
erulkey: VERCEL_TELEMETRY_DISABLED is the ecosystem-wide switch (the same var the vercel CLI documents), VERCEL_SANDBOX_TELEMETRY_DISABLED is the product-scoped one - so someone can turn off sandbox telemetry without silencing the vercel CLI's. Honoring both came out of the data-side review: one opt-out story covering both the events and this header

packages/vercel-sandbox/src/api-client/api-client.ts
erulkey: the space is there - agentSuffix is ` agent/${name}` with a leading space. Live capture: `vercel/sandbox/3.2.0 agent/claude-code_2-1-247_agent (Node.js/v26.7.0; darwin/arm64)`. I kept it as a separate product token (RFC 7231 user-agent = product tokens + optional comment) rather than inside the parens because the CLI's dedupe guard and the opt-out strip both match on ` agent/`, and a standalone token is also what the server side will parse for once ingestion lands. Happy to move it into the comment form in the shared-code follow-up if you prefer - it just needs to touch the guard, the strip and their tests together

PR #322, opened 2026-09-03
feat(sdk): add `driveId` getter on `Drive` class

QuiiBz:
The API is now returning a unique `id` field when creating or retrieving a drive, so this PR exposes it in the SDK via a new `driveId` getter

### DIFF
--- a/packages/vercel-sandbox/src/api-client/validators.ts
+++ b/packages/vercel-sandbox/src/api-client/validators.ts
@@ -270,6 +270,7 @@ export const SnapshotResponse = z.object({
 });
 
 export const Drive = z.object({
+  id: z.string(),
   name: z.string(),
   projectId: z.string(),
   region: z.string(),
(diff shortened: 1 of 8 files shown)


packages/vercel-sandbox/src/api-client/validators.ts
```diff
@@ -270,6 +270,7 @@ export const SnapshotResponse = z.object({
 });
 
 export const Drive = z.object({
+  id: z.string(),
```
marc-vercel, line 273: If it is not optional, old drives will fail when fetching them with the new SDK. Is that acceptable?

marc-vercel (approved):
LGTM, left a small comment.

packages/vercel-sandbox/src/api-client/validators.ts
QuiiBz: We always return an `id` field, it fallbacks to the drive name if there was no ID generated (for very old drives)

PR #340, opened 2026-09-18
[cli] warn when forking a running sandbox

erulkey:
Context: Gauntlet story [Fork a prepared sandbox into multiple work sandboxes](https://gauntlet.playground-vercel.tools/stories/8f2e9ffc-a515-49eb-8b2c-8f1d133f6a22). The agent created a source, wrote `/workspace/prepared/state.txt`, ran `sandbox fork` twice, and both children came up without the file. It recovered only after finding `sandbox snapshot --stop` on its own and forking again: 12 of 22 steps adapted.

The server forks from the source's latest snapshot, or from a fresh copy of its runtime when there is none. It never reads a running session's filesystem. The SDK JSDoc and the vercel.com CLI reference say so; the CLI did not. `fork --help` only mentioned copied config, and forking a running source printed the normal success summary and exited 0.

Reproduced with sandbox 4.3.0 against `elisabethrulke/sandbox-demo`:

```
$ sandbox exec fork-warn-src -- sh -c 'mkdir -p /workspace/prepared && echo ready > /workspace/prepared/state.txt'
$ sandbox fork fork-warn-src --name fork-warn-src-a
✅ Sandbox fork-warn-src-a forked from fork-warn-src.
$ sandbox exec fork-warn-src-a -- cat /workspace/prepared/state.txt
cat: /workspace/prepared/state.txt: No such file or directory
exit=1
```

This:

- looks the source up before forking (one extra GET, `resume: false` so a stopped source stays stopped) and, when its session is `running`, prints a warning to stderr with the snapshot-first workaround. The fork still proceeds.
- states the snapshot semantics in `fork --help`; `docs/index.md` regenerated.

Same sequence with this branch:

```
$ sandbox fork fork-warn-src --name fork-warn-src-b
warn: fork-warn-src is running. The fork starts from its runtime image (it has no snapshot yet), not from its live filesystem.
╰▶ hint: run `sandbox snapshot --stop fork-warn-src` first to fork the current filesystem.
- Forking sandbox fork-warn-src...
✅ Sandbox fork-warn-src-b forked from fork-warn-src.
```

After `sandbox snapshot --stop fork-warn-src`, the next fork printed no warning and `cat /workspace/prepared/state.txt` in the child returned `ready`.

Not done here: the warning fires on `running` only; `stopping` and `snapshotting` sources are left alone. I did not add a snapshot-and-fork flag, since it would stop the source as a side effect and deserves its own discussion. Server behavior is unchanged.

Verification: `pnpm vitest run test/commands` 22 passed (3 new in `fork.test.ts`), `pnpm typecheck` clean, live run above, all 4 test sandboxes removed afterwards.


packages/sandbox/docs/index.md
```diff
@@ -12,7 +12,7 @@ Commands:
     ls | list                                  List all sandboxes for the specified account and project.
     create                                     Create a sandbox in the specified account and project.
     sh                                         Create a sandbox and start an interactive shell
-    fork           <source>                    Fork an existing sandbox into a new one. Copies config (cpu, timeout, network policy, tags, env vars, etc.) from the source sandbox; any flag passed here overrides the copied value.
+    fork           <source>                    Fork an existing sandbox into a new one. The fork starts from the source's latest snapshot (or a fresh copy of its runtime when it has none) and copies its config (cpu, timeout, network policy, tags, env vars, etc.); any flag passed here overrides the copied value. Changes made in a running source since its last snapshot are not included: run `sandbox snapshot --stop <source>` first to fork the current filesystem.
```
marc-vercel: We can remove changes from this file. It will be automatically generated by changeset in the PR it will create.

packages/sandbox/src/commands/fork.ts
```diff
@@ -133,6 +133,28 @@ export const fork = cmd.command({
     const tagsObj = Object.keys(tags).length > 0 ? tags : undefined;
     const envObj = Object.keys(envVars).length > 0 ? envVars : undefined;
 
+    // The server forks from the source's latest snapshot, never from a live
+    // session, so a running source silently leaves its recent changes behind.
+    const sourceSandbox = await sandboxClient.get({
```
marc-vercel, line 138: I understand we are throwing a readable error when the sandbox is not found (fork would fail anyways).

marc-vercel (approved):

packages/sandbox/docs/index.md
erulkey: Good catch, reverted — pushed in e9a4333. This will regenerate correctly via the changeset version-bump flow.

packages/sandbox/src/commands/fork.ts
erulkey: Confirmed — `formatApiError` already special-cases a 404 on the sandbox lookup into a readable message with a `sandbox ls --all` hint, so fork fails with a clear error on a missing source.

packages/sandbox/docs/index.md
```diff
@@ -1,7 +1,7 @@
 ## `sandbox --help`
 
 ```
-sandbox 4.4.0
+sandbox 4.3.0
```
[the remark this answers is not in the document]
erulkey: Confirmed and fixed in 0355d1b — e9a4333 reverted against a stale local main ref, not origin/main, which dropped the drives/telemetry/--mount docs and downgraded the version header. Redone against origin/main; diff is now empty.

PR #342, opened 2026-09-18
[cli] ask for the drive size limit interactively and accept sizes with units

erulkey:
Follows #341. Today `sandbox drives get-or-create <name>` silently creates a 1 TiB drive, and `--max-size` only takes raw bytes: `--max-size 2TiB` is parsed as `2` and reaches the API, which rejects it with the 1 GiB minimum and never says the unit was dropped. The size is fixed after creation (a different size on re-request is a 409), so getting it wrong means deleting and recreating.

This:

- accepts sizes with units on `--max-size` (`200GiB`, `2TiB`, `5GB`; bytes if no unit) and rejects anything that does not parse:

  ```
  $ sandbox drives get-or-create chk --max-size "2 apples"
  error: found 1 error
    drives get-or-create chk --max-size 2 apples
                                        ^ Invalid size "2 apples". Use bytes or a size with a unit, e.g. 200GiB or 2TiB.
  ```

- asks a person at a terminal for the size limit when `--max-size` is omitted and the drive does not exist yet, re-asking on invalid input:

  ```
  $ sandbox drives get-or-create chk-tty
  ? Size limit for chk-tty (fixed after creation) › defaults to 1 TiB, accepts GiB or TiB, configurable up to 16 TiB on this team. Enter to continue with the default.
  › 5GiB
  ✅ Drive chk-tty ready.
     │ region: iad1
     │ max size: 5 GiB
     ╰ created: 0 seconds ago
  ```

**Agents, scripts and CI are never prompted.** The question appears only when stdin and stderr are both TTYs, `CI` is unset, `@vercel/detect-agent` (already used for telemetry) finds no agent, and the drive is not already present. Any one of those failing gives today's behaviour byte for byte: no question, API default, same output. `--max-size` always skips the prompt, and the SDK is untouched. Same build, stdin from `/dev/null`:

  ```
  $ sandbox drives get-or-create chk-default < /dev/null
  ✅ Drive chk-default ready.
     │ region: iad1
     │ max size: 1 TiB
     ╰ created: 0 seconds ago
  ```

  and under a real pty from inside a Claude Code session (`CLAUDECODE` set) the prompt is skipped the same way. The extra `GET /v2/sandboxes/drives?namePrefix=` lookup runs only inside the interactive gate, so headless runs make no additional requests.

`GB`/`TB` are treated as the binary units, because the docs only ever describe drive sizes in GiB/TiB and the API takes bytes; a decimal reading would land 7-10% under what was typed. The prompt copy uses the documented defaults; there is no API for a team's plan or quota, so on Hobby the real default is 1 GiB while the copy says 1 TiB. I'm keeping the copy and noting it here rather than guessing the plan client-side.

Verification: `pnpm vitest run` in `packages/sandbox`, 141 tests passed (121 on main, 20 new across `test/args/drive-size.test.ts`, `test/util/prompt.test.ts`, `test/commands/drives.test.ts`); `pnpm typecheck` clean; live against Internal Playground: headless default, `--max-size 5GiB`, garbage rejection, pty prompt with `5GiB`, pty prompt with `5 apples` then `2TiB` (one re-prompt), and the in-agent pty run above, all observed as pasted. Ctrl-C at the prompt rejects the pending question on Node 26 and the CLI exits 1 with the usual red line; on Node versions where `rl.question` does not reject on close this is UNVERIFIED.


packages/sandbox/src/commands/drives.ts
```diff
@@ -131,6 +153,35 @@ const getOrCreate = cmd.command({
...
+ * getOrCreate reports the real error afterwards.
+ */
+async function driveExists(params: {
+  token: string;
+  teamId: string;
+  projectId: string;
+  name: string;
+}): Promise<boolean> {
+  try {
+    const { drives } = await driveClient.list({
```
marc-vercel: We are doing a `.list()` for what is expected to be a exact name lookup. Do we have a way of querying just a single drive by name, instead of list?

If we go with list, we must ensure that the exact match always appears as the first item, and if so, we can decrease limit to 1 also.

packages/sandbox/src/util/prompt.ts
```diff
@@ -0,0 +1,56 @@
...
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stderr,
+  });
+  try {
+    for (;;) {
+      const answer = await rl.question(
+        `${chalk.cyan("?")} Size limit for ${chalk.cyan(name)} (fixed after creation) ${chalk.dim("›")} ` +
+          "defaults to 1 TiB, accepts GiB or TiB, configurable up to 16 TiB on this team. " +
+          `Enter to continue with the default.\n${chalk.dim("›")} `,
```
marc-vercel: I would try to remove from the prompt sizes and make it more concise. We can change maximum sizes on a per-team basis and for example Hobby is 1GiB.

I would just ask for the drive size in GiB, so that users only have to enter numbers and don't have to think about appending GB, GiB or whatever.

packages/sandbox/src/commands/drives.ts
erulkey: Gone in 48bef9e, along with the prompt it existed for. For the record there's no single-drive GET on the API surface (list, get-or-create, delete only), and I had pinned `sortOrder: asc` so the exact name always landed on page one before removing the lookup entirely.

packages/sandbox/src/util/prompt.ts
erulkey: Prompt is removed, no customer had asked for it and you're right that the CLI has always used flags. Your per-team point is exactly why the replacement states no numbers: when `--max-size` is omitted the line now reads `max size: 1 TiB (set with --max-size at creation, fixed afterwards)`, which points at the flag without claiming a default or a ceiling, so it stays correct on Hobby and on teams with a raised quota. It also avoids lying on an existing drive, since `get-or-create` returns whatever size that drive was created with and the response has no created-vs-retrieved signal. What's left in the PR is the unit parsing on `--max-size` plus rejecting input it can't parse.

marc-vercel (approved):

